PR #3, opened 2016-08-15
Added missing API methods and documentation. Removed internall HidDevice buffer.

newpavlov:
Additionally added example of reading data from KIT MT 8057 CO2 detector. Fixed some clippy notifications.      Made HidDeviceInfo's fields public, removed getters.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "hidapi"
-version = "0.2.0"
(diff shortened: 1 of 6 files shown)


newpavlov:
Also I thought of making all of `HidDeviceInfo`'s fields public and removing `impl HidDeviceInfo`. (in my opinion set of getters in this case is quite redundant) But removal of internal buffer is breaking enough change, so I didn't want to introduce more of it without discussing it first. :)

PR #13, opened 2017-12-23
Use hidraw instead of libusb backend on Linux

drola:
While testing hidapi-rs on Ubuntu I found that:
 - I need to run software as a superuser to be able to connect to devices
 - After my application exists computer stops responding to keyboard and mouse if I connect to them through hidapi-rs

After a bit of research I found that libusb implementation of hidapi disconnects kernel's hidraw driver when opening a HID device. Additionally it is not possible to access HID devices connected through Bluetooth. Switching to hidraw backend for hidapi fixes issues mentioned above.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,12 @@ build = "build.rs"
 links = "hidapi"
 documentation = "https://docs.rs/hidapi"
 
+[features]
+# Enable this feature to use hidraw backend instead of
+# libusb-1.0 on Linux. Enabling this feature does not
+# affect other platforms.
+linux-hidraw = []
+
 [dependencies]
 libc = "0.2.15"
 
--- a/README.md
+++ b/README.md
@@ -13,6 +13,18 @@ used by adding `hidapi` to the dependencies in your project's `Cargo.toml`.
 [dependencies]
 hidapi = "0.3"
 ```
+
+On Linux you can choose between two hidapi backends: `libusb-1.0` or `hidraw`.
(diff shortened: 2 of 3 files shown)


ruabmbua:
@drola you can always add a rule to /etc/udev/rules.d , which changes access permissions to your HID device. I added a group for my keyboard backlight, which has access to the device.

Edit:
Also you should probably not just switch the linux build to hidraw. You can for example add config flags / feature sets to cargo, and let the library user choose the backend implementation.

drola:
My use case requires using hidraw driver because I am implementing advanced functionality on Logitech G203 gaming mouse. I want the kernel to normally handle mouse pointer while simultaneously setting and getting some vendor-specific reports from the device.

Letting the end user choose implementation would be the best option but unfortunately I lack the knowledge to implement it that way. It looks like [specifying features in Cargo.toml](http://doc.crates.io/manifest.html#the-features-section) is the way to go? I could look into this tomorrow and update this PR.

ruabmbua:
Yes features are the correct way. When you add some #[cfg(...)] magic in your build.rs you can even turn on hidraw by default on linux, and let it override with a feature flag.

drola:
I updated the pull request.

By default it will use `libusb-1.0` backend on Linux to preserver current behaviour. Optionally a feature called `linux-hidraw` can be enabled to switch to `hidraw` backend. I added `Cargo.toml` example to the Usage section in `README.md`.

ruabmbua:
Closed in favor of #12 .

PR #14, opened 2017-12-31
Correct docs for HidDevice::get_feature_report

phrohdoh:
Previously the docs mentioned a non-existent `data` parameter.
This was likely a copy and paste blunder.


ruabmbua:
@Phrohdoh thanks for the PR, I merged it now. Sorry it took so long, but I have actually taken over as crate maintainer just now.

PR #18, opened 2018-02-15
Fix the build script when cross-compiling

tomaka:
The target we're compiling to is not necessarily the same as the one of the build script.


ruabmbua:
Closed in favor of #12 .

PR #25, opened 2018-08-05
Update README.md

jleni:
The example needs to be adjusted when using version 0.5.0


ruabmbua (commented):
Thanks, I forgot that. Doctests in readmes would be great.

PR #35, opened 2019-02-23
fix(replace `static mut` with `static AtomicBool`)

niklasad1:
Hey, this replaces `global static mut` with `AtomicBool` which is thread-safe


ruabmbua:
Thanks for the contribution! You probably figured out, that the whole API is not threadsafe (does not implement Sync), but still, the acquiring of the API guard should be threadsafe.

PR #36, opened 2019-02-26
Fix incorrect compare_and_swap check

jleni:
At the moment (v0.5.1) valid calls to `acquire` fail. Any application using this library will fail to acquire an API handle.

`compare_and_swap` returns the previous value so in this case a valid lock results into an `InitializationError`


ruabmbua:
Thanks!

Issue #16, opened 2018-01-07
Trying to store a hidapi::HidDevice in a struct


shayne:
Hey there, thanks for the library!

I'm working on a project to abstract away a device. So simply put, I'm looking to store the `hidapi::HidDevice` that I open in a struct so my API can call `read/write` on it as necessary.

It's easy enough to store the api, but once I try to store the device (alone or with api) I get lifetime issues. Perhaps this has to do with the `PhantomData` you have on `HidDevice`?

I'm probably just missing something with Rust but I've been struggling to abstract away the HidApi from my library.

ruabmbua:
Hi!
Take a look at this PR, which will solve your problem. I ran into the same issue a while ago.
Link: https://github.com/Osspial/hidapi-rs/pull/12

The PR is not yet merged, but you can always add my branch as a cargo git dependency if you want.

shayne:
This is awesome, thanks for the quick response. I'll use your branch for now- glad I wasn't the only one who felt like it was a pain in the ***.

shayne:
Hey @ruabmbua how are you sending the device between threads? Your PR mentions it but getting all sorts of errors.

ruabmbua:
The HidDevice / API have to implement the Send trait, to be able to be shared safely between threads. My branch does not do this, and only solves the "store inside same struct" and "lifetime" issue.

My program is multi threaded, but relies on the gtk event queue, which solves synchronization issues for me.

When you want to access the API from multiple threads, try to spawn a handler thread, which owns / manages all your devices, and then pass std::sync::mpsc channels to it. You can then send your own custom messages from and to the hid IO thread safely.

ruabmbua:
Tip: You can check if your type / instantiation can be shared between thready by doing this:

Remove the comments, to see the compilation errors for types, which do not implement Send. 

std::rc::Rc does not implement send, because it has a non threadsafe reference counter inside. This is also true for all types, which embed the Rc type. There are also other types which do not implement send, like Raw pointers.

https://play.rust-lang.org/?gist=81b3dfd6d61db891b55fac499b354fe0&version=stable

shayne:
Super helpful again - glad my thinking agreed with your assessment. Currently building out my final API using channels.

ruabmbua:
@shayne finally my fork is merged now. I will release a new version to crates.io sometime soon.

shayne:
Awesome thanks! I'll look at updating my end soon enough

PR #39, opened 2019-05-06
Made HidApi struct thread-safe

AlexCouch:
I ran into a problem where I had different threads trying to reference my program's HidApi instance. However, HidApi was not thread-safe due to the _lock field being a non-async reference counted HidApiLock `Rc<HidApiLock>`. I changed this to be `Arc<HidApiLock>`, that way multiple threads can reference the hidapi instance. I provided a really crappy example of it working.

I know the underlying lower level implications of this that this could allow dependents to write to a HidDevice from multiple threads but I think that we cannot control that due to the fact that this relies on the C implementation of hidapi as the backend. I see that there is currently plans for isolating this library into its own pure rust implementation and I believe that would make way for a more thread-safe implementation of this library unless we can do some kind of hacky write access block if the device is currently already being written to in the write binding. I haven't completely explored this yet but I honestly just needed a _simple_ way of allowing multiple threads to reference my hidapi instance. My project's implementation is simply just a single thread for reading and a single thread for writing. This is the fix that I came up with.

I know the example is very bad and I couldn't demonstrate multiple device opens in parallel but it seems to be just fine. And that's another thing that could be dangerous: multiple threads opening a device concurrently. The point is, I made this library just a tiny bit more thread-safe while also introducing some possible hazards. But it's a fix for now I guess.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,3 +31,6 @@ failure_derive = "0.1.2"
 [build-dependencies]
 cc = "1.0.3"
 pkg-config = "0.3.9"
+
+[dev-dependencies]
+threadgroup = "0.1.0"
--- a/examples/concurrency_test.rs
+++ b/examples/concurrency_test.rs
@@ -1,0 +1,97 @@
+extern crate hidapi;
+extern crate threadgroup;
+extern crate core;
+
+use hidapi::{HidApi, HidResult};
+use std::sync::{Arc, Mutex};
+use std::thread;
+
+struct AnotherTestThing;
+
+impl AnotherTestThing{
+    fn do_something(test_thing: Arc<TestThing>){
+        thread::spawn(move || {
+            let test_clone1 = Arc::clone(&test_thing);
+            let thread1 = thread::Builder::new().name("thread1".to_string()).spawn(move || {
+                let hidapi = Mutex::lock(&test_clone1.hidapi).unwrap();
+                let id = thread::current().id();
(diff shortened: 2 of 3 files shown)


ruabmbua:
Hi!

Yes you are indeed correct, that just making the lock threadsafe does not give you any real thread safety.

I am however not quite sure, if signal11/hidapi is or is not threadsafe.

Someone else had problems with accessing hidapi with multiple threads, and I suggested using a channel based API: https://github.com/ruabmbua/hidapi-rs/issues/16

You will have to read through the whole thread + comments, somewhere in the middle of it the channel API is mentioned.

There is also https://docs.rs/hidapi/0.5.2/hidapi/struct.HidDevice.html#method.set_blocking_mode. You can set it to false, to activate non blocking mode. If everything you want is fast reaction time, then you can set up a event loop, where you handle your protocol. Usually you would use something like *poll*, *epoll*, etc..., but that is not really supported by this crate.

Example (pseudo rust code):

```rust

fn io_loop() {
    let mut device = ...;
    device.set_blocking_mode(false);

    loop {
         let rdata = device.read(...); // Will return immediately
         handle_read_data(rdata);
         if check_if_want_write() {
             device.write(...);
         }
         // Optionally put something like thread::sleep(...); here to not burn all your cpu time
    }
} 
```

ruabmbua:
There is also HidDevice::read_timeout(). It will guarantee to return, when the given timeout is over.

PR #43, opened 2019-07-24
Implement Error manually instead of using failure

### LINKED ISSUE #42: Consider dropping the failure depedency
The `failure` dependency doesn't do much here, and is a bit heavy because of the transitive dependency to `syn`. Removing it reduces the release build time from 13 to 4 seconds.

`failure` also brings some interoperability problems with crates that don't use it.

I can submit a PR to replace it with a hand-rolled implementation, but it's technically a breaking change. Do you have anything against it? And if we introduce a breaking change now, can we also update to the 2018 edition and start using `dyn Trait` to get rid of some compiler warnings?

lnicola:
Closes #42.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "hidapi"
-version = "0.5.2"
+version = "0.6.0"
 authors = [
     "Roland Ruckerbauer <roland.rucky@gmail.com>",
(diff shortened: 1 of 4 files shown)


ruabmbua (approved):
Does not really add many LoC, and decreases compile time, thanks!

PR #44, opened 2019-07-28
Allow HidDevice to be sent across threads

mon:
As far as I can tell this is safe (despite the empty `unsafe` impl for `Send`), since it's still using the lock to control API access.

This provides a really nice usability boost for me as I can integrate hidapi with rocket for making web interfaces for devices.


ruabmbua (approved):
Just implementing Send looks fine

PR #49, opened 2019-12-02
Add FreeBSD support

neverpanic:
Introduce a method to find a copy of hidapi using pkg-config on FreeBSD. This currently only supports shared linking.


ruabmbua (approved):

ruabmbua:
Thanks!

Issue #46, opened 2019-08-04
The hidapi repository is no longer maintained


tim-hellhake:
According to https://github.com/signal11/hidapi/issues/431 the development was moved to https://github.com/libusb/hidapi.

ruabmbua:
Solved

PR #50, opened 2020-01-23
Update hidapi to use libusb fork

haata:
- signal11 repo has not been maintained since 2016
- See Issue #46


ruabmbua:
Thanks, but I will update the submodule myself, and merge against remote master.

Issue #52, opened 2020-02-29
Can't create a HidApi instance on Windows 7


fpagliughi:
Hi. I'm trying to build and run the example apps on Windows 7, and am getting an error when it tries to calls `HidApi::new()`:

```
./target/debug/examples/lshid.exe
Printing all available hid devices:
Error: failed converting 0xDDCD to rust char

./target/debug/examples/lshid.exe
Printing all available hid devices:
Error: failed converting 0xDEE7 to rust char
```

Any ideas?

I'm using the latest Rust 1.41.1 with hidapi v 1.1.1

ruabmbua:
You probably have a device attached, which contains some invalid UTF-16 in e.g. the serial number, product name etc...

You can handle the `HidError::FromWideCharError` for your case.

ruabmbua:
Update: I think it is impossible currently to enumerate devices, when one contains some invalid UTF-16...

Did not expect that one. I think I can come up with a fix. Should not take long. Are you willing to test, if it fixes your problem?

lnicola:
I wonder if the device actually returns an invalid string, or something along the way (Windows) messes it up.

fpagliughi:
Yes. Interesting. I hit a bug in the _rusb_ crate with this same machine a few days ago:
https://github.com/a1ien/rusb/issues/15

It seemed that a device was returning at least one bad descriptor claiming a length of 255 bytes and very suspect contents. In that case the fix that I submitted did a basic check and didn't even try to parse the string in that case. For example... if it returns a string with an odd number length, then that is definitely wrong. Don't even try to parse it. Return an error.

fpagliughi:
Sorry I needed to step away from the computer for a short while. I will be back in a few minutes to test the update and report back.

fpagliughi:
Well it doesn't crash anymore, which is good, but now I'm seeing this:
```
lshid.exe
Printing all available hid devices:
VID: 046a, PID: 010d, Serial: <COULD NOT FETCH>, Product name: <COULD NOT FETCH>
VID: 046a, PID: 010d, Serial: <COULD NOT FETCH>, Product name: <COULD NOT FETCH>
VID: 046d, PID: c52f, Serial: <COULD NOT FETCH>, Product name: ∁CԇंĄĀ
VID: 046d, PID: c52f, Serial: <COULD NOT FETCH>, Product name: ࠀ
[...]

PR #53, opened 2020-02-29
Do not fail on wchar_t* to string conversion

ruabmbua:
Quick fix for issue #52.

Instead of failing, return None value.

Edit:

Also added new DeviceInfo with access to raw wchar_t strings, and deprecated HidDeviceInfo and friends.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "hidapi"
-version = "1.1.1"
+version = "1.2.0"
 authors = [
     "Roland Ruckerbauer <roland.rucky@gmail.com>",
     "Osspial <osspial@gmail.com>",
--- a/examples/lshid.rs
+++ b/examples/lshid.rs
@@ -17,10 +17,14 @@ fn main() {
         Ok(api) => {
             for device in api.device_list() {
                 println!(
-                    "VID: {:04x}, PID: {:04x}, Serial: {}",
+                    "VID: {:04x}, PID: {:04x}, Serial: {}, Product name: {}",
                     device.vendor_id(),
                     device.product_id(),
                     match device.serial_number() {
+                        Some(s) => s,
+                        _ => "<COULD NOT FETCH>",
+                    },
+                    match device.product_string() {
                         Some(s) => s,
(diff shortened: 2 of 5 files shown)


ruabmbua:
@fpagliughi Can you try this fix?

lnicola:
Should these be `OsString`s instead?

ruabmbua:
@lnicola Maybe. The problem is, OsString claims, it is the OS representation of strings, but the wchar_t* in this case are HID / USB specific, so they don`t clearly match OsString. I could however export an interface with direct access to the wchar_t* strings.

What do you think?

lnicola:
Ah, I didn't realize that. https://docs.rs/widestring/0.4.0/widestring/ might be a solution here?

> All four string types may still have unpaired UTF-16 surrogates or invalid UTF-32 code points; ill-formed data is preserved until conversion to a basic Rust String.

ruabmbua:
Well, currently I use char::from_u32(), which should convert from any valid unicode codepoint. If it fails, its just not a valid codepoint. widestring won`t be able to do anything about that.

ruabmbua:
> Well, currently I use char::from_u32(), which should convert from any valid unicode codepoint. If it fails, its just not a valid codepoint. widestring won`t be able to do anything about that.

Ups, did not read your citation...

Still, I`d like to leave this to the user. I will add access to the underlying representation.

ruabmbua:
Ahh all stuff in HidDeviceInfo is public... I do not want to break compatibility...

lnicola:
I figured. We could use `widestring` now internally and do the string conversion (with `to_string` or `to_string_lossy`) and either expose separate `widestring` methods now (like `env` vs. `env_os`) or make the change in the next breaking version.

But maybe you think it's too much of a hassle?

ruabmbua:
I think I will deprecate `Hidapi::devices()`, and make a new one which returns a iterator with a new DeviceInfo type. This time with getters.

ruabmbua:
Ok, I added DeviceInfo as an alternative.

Its somewhat ugly internally (keeping around a compatibility, and new list). Can be removed after next major breaking change.

PR #70, opened 2021-04-28
Static CRT support

### LINKED ISSUE #69: Static CRT option for Windows
It would be nice if this crate allowed to link C runtime library statically, as a cargo feature option (say, `windows-static-crt`).

https://github.com/ruabmbua/hidapi-rs/blob/06354516eb65a65319748c28ea622b3eb324005a/build.rs#L115-L121

 needs to be modified to include a call to `static_crt(true)` 

See https://docs.rs/cc/1.0.67/cc/struct.Build.html#method.static_crt

pravic:
Tested, works fine.

It builds static CRT only when required: https://doc.rust-lang.org/reference/linkage.html#static-and-dynamic-c-runtimes.

Fixes #69.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -113,12 +113,14 @@ fn compile_freebsd() {
 }
 
 fn compile_windows() {
+    let linkage = env::var("CARGO_CFG_TARGET_FEATURE").unwrap_or(String::new());
+
     let mut cc = cc::Build::new();
     cc
         .file("etc/hidapi/windows/hid.c")
         .include("etc/hidapi/hidapi");
 
-    if cfg!(target_feature = "crt-static") {
+    if linkage.contains("crt-static") {
         // https://doc.rust-lang.org/reference/linkage.html#static-and-dynamic-c-runtimes
         cc.static_crt(true);
     }
--- a/build.rs
+++ b/build.rs
@@ -116,6 +116,7 @@ fn compile_windows() {
     cc::Build::new()
         .file("etc/hidapi/windows/hid.c")
         .include("etc/hidapi/hidapi")
+        .static_crt(true)
         .compile("libhidapi.a");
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -113,10 +113,16 @@ fn compile_freebsd() {
...
 fn compile_windows() {
-    cc::Build::new()
+    let mut cc = cc::Build::new();
+    cc
         .file("etc/hidapi/windows/hid.c")
-        .include("etc/hidapi/hidapi")
-        .compile("libhidapi.a");
+        .include("etc/hidapi/hidapi");
+
+    if cfg!(target_feature = "crt-static") {
```
ruabmbua: I think it is not valid to use *cfg!* macro or the *cfg* attribure for this, it may break cross compilation. build.rs compilation is detached from the main library build.

You can access the *target_feature* of the library from the build.rs with the *CARGO_CFG_TARGET_FEATURE* environment variable.

See e.g. https://github.com/google/shaderc-rs/blob/647e07510c160f6d15e63edd6c4f089e353cef76/shaderc-sys/build/build.rs#L84 for an example.

ruabmbua (changes requested):

build.rs
pravic: > cross compilation

Hmm.. I saw both ways, actually. You're using the env variable, I know. I thought, they are interchangeable. Is there any link to dig about cross-compilation in cargo?

pravic: Check db2065a

ruabmbua: I am sure, it is mentioned somewhere in the reference.
If you think about it, the build.rs can not really knowm which cfgs are enabled with cfg!, because that is resolved at compile time of the build.rs. The needed information is however passed into the resulting build.rs executable at runtime.

ruabmbua:
Thank you for implementing this!

PR #72, opened 2021-11-14
Retrieve error message from hidapi for general errors

Tiwalun:
The `hid_error` function can also be used when opening a device.


ruabmbua:
Thank you!

PR #73, opened 2021-11-21
add illumos support

jclulow:
Works with either a system libusb or a system libusb-hidapi found
through pkg-config.


jclulow:
I've tested that both of the new features build and link against the appropriate libraries, that `cargo test` works, and that the `lshid` example appears to execute once built.

ruabmbua:
Thank you for the contribution!

PR #74, opened 2021-12-18
feat: add linux-static-rusb as backend

joshieDo:
Adds rusb as backend, so we don't depend on `libusb` being installed on the machine, or are forced to use another backend (and thus required to have `libudev` installed)

I copy-pasted the `libusb.h` file from the `rusb` repository.  Should we add `rusb` / `libusb1-sys` as submodule and link it from there, or does this suffice?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@ language: rust
 
 matrix:
   include:
+    - env: TARGET=x86_64-unknown-linux-gnu FEATURE_FLAGS="linux-static-rusb"
+
     - env: TARGET=x86_64-unknown-linux-gnu FEATURE_FLAGS="linux-static-libusb"
 
     - env: TARGET=x86_64-unknown-linux-gnu FEATURE_FLAGS="linux-static-hidraw"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ documentation = "https://docs.rs/hidapi"
 default = ["linux-static-libusb", "illumos-static-libusb"]
 linux-static-libusb = []
 linux-static-hidraw = []
+linux-static-rusb = ["rusb"]
 linux-shared-libusb = []
 linux-shared-hidraw = []
 illumos-static-libusb = []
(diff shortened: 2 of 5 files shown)


ruabmbua:
First, thanks for the PR, however:

I am not sure, if adding rusb is a good idea, it would not really be used for its actual functionality (providing a rust wrapper for libusb). Also not sure, how big the benefit is, to actually build libusb, instead of using the system installed library.

If you really need such functionality, I propose a patch to hidapi-rs, where the build script does not automatically link to libusb any more. Then, your project / crate can still bring in libusb via rusb, and link to it via its build.rs.

joshieDo:
> Also not sure, how big the benefit is, to actually build libusb, instead of using the system installed library.

The goal would be having all dependencies accessible from one single "context", in this case `cargo`. But I understand what you mean. And since it would be feature-based, i thought it would be okay.

>If you really need such functionality, I propose a patch to hidapi-rs, where the build script does not automatically link to libusb any more. Then, your project / crate can still bring in libusb via rusb, and link to it via its build.rs.

To make sure I understand, do you mean this below, and then linking on my own crate both `libhidapi` (since it requires <libusb.h> to compile) and `libusb`? Or did I misunderstand?

```diff
-                let lib =
-                    pkg_config::find_library("libusb-1.0").expect("Unable to find libusb-1.0");
-                for path in lib.include_paths {
-                    config.include(
-                        path.to_str()
-                            .expect("Failed to convert include path to str"),
-                    );
-                }
-                config.compile("libhidapi.a");
+
+                match pkg_config::find_library("libusb-1.0")
+                {
+                    Ok(lib) => {
+                        
+                        for path in lib.include_paths {
+                            config.include(
+                                path.to_str()
+                                    .expect("Failed to convert include path to str"),
+                            );
+                        }
+                        config.compile("libhidapi.a");
+                    },
+                    Err(_) => {
+                        println!("Unable to find libusb-1.0. Link libhidapi and libusb on your own!")
+                    },
+                };
```

Thank you!

ruabmbua:
Yes, exactly. Although I would by default fail to build, and feature-gate the soft error behind e.g. a `linux-custom` backend.

PR #79, opened 2022-03-05
Add OpenBSD to supported platforms

VlkrS:
This PR adds OpenBSD to the supported platforms. 

Unfortunately hidapi's `hid_init()` needed to be renamed to `hidapi_hid_init()` on OpenBSD due to a naming conflict with a system library function, so we need to work around that, too. I still hope you'll consider it for inclusion.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -70,6 +70,12 @@ impl HidApiLock {
         {
             // Initialize the HID and prevent other HIDs from being created
             unsafe {
+                #[cfg(target_os = "openbsd")]
+                if ffi::hidapi_hid_init() == -1 {
+                    HID_API_LOCK.store(false, Ordering::SeqCst);
+                    return Err(HidError::InitializationError);
+                }
+                #[cfg(not(target_os = "openbsd"))]
                 if ffi::hid_init() == -1 {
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -70,6 +70,12 @@ impl HidApiLock {
         {
             // Initialize the HID and prevent other HIDs from being created
             unsafe {
+                #[cfg(target_os = "openbsd")]
+                if ffi::hidapi_hid_init() == -1 {
```
ruabmbua: Instead of copying this block, either create an unsafe wrapper function around the hid_init functions, or alternatively keep only *hid_init*, and use https://doc.rust-lang.org/reference/items/external-blocks.html#the-link_name-attribute with *cfg_attr* to just rename the symbol. Then you do not need to use *cfg* here.

ruabmbua (changes requested):

src/lib.rs
VlkrS: Done. Diff looks much cleaner now.

ruabmbua: Nice, thank you!

ruabmbua:
Thanks for the contribution, new version is already published to crates.io.

PR #80, opened 2022-03-31
Add wrap sys device

mukuact:


src/lib.rs
```diff
@@ -91,6 +91,43 @@ impl Drop for HidApiLock {
     }
 }
 
+/// Open a HID device using libusb_wrap_sys_device.
+/// This cannot use HdiApi because new() is failed.
```
ruabmbua: Please clarify what exactly this means, I am not quite sure.

src/lib.rs
```diff
@@ -91,6 +91,43 @@ impl Drop for HidApiLock {
...
+            message: unsafe {
+                match wchar_to_string(ffi::hid_error(std::ptr::null_mut())) {
+                    WcharString::String(s) => s,
+                    _ => return Err(HidError::HidApiErrorEmpty),
+                }
+            },
+        });
+        return e;
+    }
+    let lock = HidApiLock::acquire()?;
```
ruabmbua: Looks like a bad idea to init hidapi and acquire lock here, makes it impossible to instantiate a safe instance of hidapi after calling this.

ruabmbua (commented):
I would prefer, if `wrap_sys_device()` would become a method on the `HidApi` struct, to proof, that hidapi is already initialized. Of course the acquire() has to be removed, in the wrap function to keep everything working.

If you want to prevent enumerating devices when creating `HidApi::new()`,  just add an alternate `HidApi::new_without_enumerate()` or similar, which bypasses the device enumeration.

src/lib.rs
```diff
@@ -91,6 +91,43 @@ impl Drop for HidApiLock {
...
+        });
+    }
+
+    let device = unsafe { 
+        ffi::libusb_set_option(std::ptr::null_mut(), 2);
+        ffi::hid_libusb_wrap_sys_device(sys_dev, interface_num) 
+    };
+
+    if device.is_null() {
+        println!("device null");
```
ruabmbua: do not printf errors in a library.
Makes it unusable for certain kind of applications.

mukuact:
wrap_sys_device became a method of HidApi, and add new initialization method.

src/lib.rs
```diff
@@ -255,6 +237,40 @@ impl HidApi {
         }
     }
 
+    /// Open a HID device using libusb_wrap_sys_device.
+    pub fn wrap_sys_device(&self, sys_dev: isize, interface_num: i32) -> HidResult<HidDevice> {
+        if cfg!(not(any(
```
ruabmbua: Instead of throwing an error at runtime, just `cfg()` this method to make it available to libusb based backend only.

b.t.w. maybe also check if this libusb functionality is also available on other platforms?

src/lib.rs
```diff
@@ -154,6 +116,26 @@ impl HidApi {
...
 
+    /// Initializes the hidapi.
+    /// it skips device scanning.
+    pub fn new_without_enumerate() ->HidResult<Self> {
+        unsafe {
+            //Do not scan for devices in libusb_init()
+            //Must be set before calling it.
+            //This is needed on Android, 
+            //where access to USB devices is limited
+            ffi::libusb_set_option(std::ptr::null_mut(), 2);
```
ruabmbua: Only invoke this, when libusb backend is used.

src/lib.rs
```diff
@@ -218,6 +238,40 @@ impl HidApi {
...
+                message: String::from("wrap_sys_device is not implement"),
+            });
+        }
+
+        let device = unsafe {
+            ffi::hid_libusb_wrap_sys_device(sys_dev, interface_num)
+        };
+
+        if device.is_null() {
+            let e = Err(HidError::HidApiError {
```
ruabmbua: call the check_error() method instead, it does all that already for you.

match on its return, and then you can return both the Ok() and Err() variants as Err()

ruabmbua (changes requested):

src/lib.rs
mukuact: If this is available to libusb based backend only, library users who use this function also have to use cfg() in their codes.
So, I think it is better to just fail the function if other backends is used, isn't it?

mukuact: Libusb wrap_sys_device functionality is available on other platforms, but  the hidapi does not use libusb as a backend on win / macOS.
So, freeBSD, openBSD can use this.
However, build.rs setting for freeBSD/openBSD does not use features.

ruabmbua: > Libusb wrap_sys_device functionality is available on other platforms, but the hidapi does not use libusb as a backend on win / macOS. So, freeBSD, openBSD can use this. However, build.rs setting for freeBSD/openBSD does not use features.

You can just use `cfg(target_os = "freebsd)", and similar for openbsd.

ruabmbua: > If this is available to libusb based backend only, library users who use this function also have to use cfg() in their codes. So, I think it is better to just fail the function if other backends is used, isn't it?

If you compile the function conditionally, the user of the library can use cfg(), and e.g. still opt into returning an error to the user at runtime. Also, the function is so specific to libusb, I do not really want to expose it to anyone else. I always prefer compile time errors over runtime.

src/lib.rs
```diff
@@ -117,6 +117,31 @@ impl HidApi {
         })
     }
 
+    /// Initializes the hidapi.
+    /// it skips device scanning.
+    pub fn new_without_enumerate() ->HidResult<Self> {
+        if cfg!(any(
```
ruabmbua: This does not actually work, the calls to libusb are still compiled, meaning this breaks all platforms except libusb based ones. See CI run.

Solve it like in this example, where the code is not actually compiled: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=18dd84fabb7e4531d0d7e0d93031be9f

src/lib.rs
```diff
@@ -218,6 +243,35 @@ impl HidApi {
...
+            feature = "linux-static-libusb",
+            feature = "linux-shared-libusb"
+        ))) {
+            return Err(HidError::HidApiError {
+                message: String::from("wrap_sys_device is not implement"),
+            });
+        }
+
+        let device = unsafe {
+            ffi::hid_libusb_wrap_sys_device(sys_dev, interface_num)
```
ruabmbua, line 247: does not actually compile successful on platforms other than libusb based ones. The `ffi::hid_libusb_wrap_sys_device()` only exists on those. Compile this whole function conditionally, to solve the problem.

src/lib.rs
mukuact: I see. I totaly misunderstood the feature system.

ruabmbua:
Thanks for the PR!

Will release a new version of the library to crates.io, after I change some minor stuff.

PR #93, opened 2022-11-11
Implement MacOS and Windows APIs from `hidapi` 0.12 release

KarlK90:
This PR implements FFI bindings and safe implementations for the following APIs that where added in `hidapi` release 0.12

**MacOS:**
  * `hid_darwin_get_location_id()`
  * `hid_darwin_set_open_exclusive()`
  * `hid_darwin_get_open_exclusive()`
  * `hid_darwin_is_device_open_exclusive()`
 
**Windows:**
   * `hid_winapi_get_container_id()`

The `hid_darwin_set_open_exclusive()` API is also put to use by introducing the `macos-shared-device` feature flag for MacOS so it is possible to have multiple `HidDevice` handles that access the same physical device.

A convenience addition is the implementation of `Debug` for `HidDevice`, this will only print `HidDevice` but allows wrapping types like `HidResult<HidDevice>` into a `dbg!` macro.

### DIFF
--- a/src/macos.rs
+++ b/src/macos.rs
@@ -4,8 +4,6 @@ use crate::ffi;
 use crate::{HidApi, HidDevice, HidResult};
 
 impl HidApi {
-    /// **`Only available on MacOS`**
-    ///
     /// Changes the behavior of all further calls that open a new [`HidDevice`]
     /// like [`HidApi::open`] or [`HidApi::open_path`]. By default on Darwin
     /// platform all devices opened by [`HidApi`] are opened in exclusive mode.
@@ -17,8 +15,6 @@ impl HidApi {
         unsafe { ffi::macos::hid_darwin_set_open_exclusive(exclusive as c_int) }
     }
 
-    /// **`Only available on MacOS`**
-    ///
     /// Get the current opening behavior set by [`HidApi::set_open_exclusive`].
     pub fn get_open_exclusive(&self) -> bool {
         unsafe { ffi::macos::hid_darwin_get_open_exclusive() != 0 }
@@ -26,8 +22,6 @@ impl HidApi {
 }
 
 impl HidDevice {
-    /// **`Only available on MacOS`**
-    ///
     /// Get the location ID for a [`HidDevice`] device.
     pub fn get_location_id(&self) -> HidResult<u32> {
         let mut location_id: u32 = 0;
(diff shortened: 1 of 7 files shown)


src/macos.rs
```diff
@@ -0,0 +1,65 @@
...
+
+impl HidDevice {
+    /// **`Only available on MacOS`**
+    ///
+    /// Get the location ID for a [`HidDevice`] device.
+    pub fn get_location_id(&self) -> HidResult<u32> {
+        let mut location_id: u32 = 0;
+
+        let res =
+            unsafe { ffi::macos::hid_darwin_get_location_id(self._hid_device, addr_of_mut!(location_id)) };
```
ruabmbua: There is no reason to use addr_of_mut!()? Could just obtain a mutable reference to the variable.

src/windows.rs
```diff
@@ -0,0 +1,35 @@
...
+    pub fn get_container_id(&self) -> HidResult<GUID> {
+        unsafe {
+            let mut container_id: GUID = std::mem::zeroed();
+
+            let res = ffi::windows::hid_winapi_get_container_id(
+                self._hid_device,
+                addr_of_mut!(container_id),
+            );
+
+            if res == -1 {
```
ruabmbua: Does not have to be in unsafe block, prefer smaller isolated unsafe blocks. (even if there have to be two)

ruabmbua (changes requested):
Thanks, LGTM

With some notes though, but easy to fix.

KarlK90:
I've added docs.rs specific metadata and documentation hints in the latest commit and also hid the `cfg_libusb_only` macro which should be internal(?). The result will look like this:

image

I find that much nicer, although there are now 2 distinct `Impl HidApi` blocks in the documentation...

src/windows.rs
KarlK90: Done! You can see the [changes here.](https://github.com/ruabmbua/hidapi-rs/compare/02a62044aaa72eeea404f30415fc6f7fd3d50e18..4dac1a20d7b4be3df307040a980a4abdbb21a87b)

src/macos.rs
KarlK90: Done! You can see the [changes here.](https://github.com/ruabmbua/hidapi-rs/compare/02a62044aaa72eeea404f30415fc6f7fd3d50e18..4dac1a20d7b4be3df307040a980a4abdbb21a87b)

KarlK90:
@ruabmbua thanks for the swift review, applied changes and added IMHO better platform specific documentation.

ruabmbua:
Thanks for the contribution!

PR #103, opened 2023-03-05
Lazily initialize hidapi.

Dirbaio:
This allows creating multiple `HidApi` instances at the same time. hidapi is initialized when the first one is created, and never deinitialized.

This issue seems to imply that hidapi is thread safe as long as you don't access the same device from 2 threads at the same time, which the current Rust bindings already enforce: https://github.com/signal11/hidapi/issues/72

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -31,7 +31,7 @@ impl Display for HidError {
                 write!(f, "failed converting {:#X} to rust char", wide_char)
             }
             HidError::InitializationError => {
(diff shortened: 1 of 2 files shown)


ruabmbua:
LGTM!

ruabmbua:
Wait, looks like there is a problem (I really should add more CI jobs).

wrap_sys_device() is not compiling when using libusb backend.

Don`t worry I fix it myself

PR #105, opened 2023-04-04
Rebuild if the hidapi file changes

carlosmn:
It's a bit of an edge case but I find myself messing around with the C code and it's annoying that cargo doesn't know to recompile.

Though I expect it would also be useful if there's an update to the submodule and you're rebuilding.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -51,6 +51,7 @@ fn compile_linux() {
             "LINUX_STATIC_HIDRAW",
             Box::new(|| {
                 let mut config = cc::Build::new();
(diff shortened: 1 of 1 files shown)


ruabmbua:
Hm, this should actually already work correctly, the cc library should already emit this cargo metadata...
I had this issue myself recently, not sure why its not working any more.

ruabmbua:
Actually, I think I am wrong... The library does not seem to detect file changes, only environment changes are automatically picked up.

ruabmbua:
Thanks for the contribution!

PR #107, opened 2023-04-06
Add a native Rust Linux hidraw backend to replace hidapi

carlosmn:
I've been meaning to get around to this and just had a few free days so here we are. My ultimate goal is being able to have something on which we can do sensible async. At worst maybe we can add a `AsRawFd` but we can probably do better.

I've been away from the device I normally use so I haven't been able to test the IO portion itself, but I have checked the enumeration shows the same via the `lshid` example including the page and usage page values. The actual IO should be trivially correct as we're just asking the kernel to do the IO (though the report get situation with the buffer is a bit suspicious). I've tried to keep the behaviour the same as the C hidapi by using `nix` to capture the system calls more accurately instead of letting the Rust stdlib make it nicer for us.

I've added this as a new feature/backend in case there are particularities of this implementation that means you might want to go back to hidapi. It also made it easier to check both backends build and enumerate the same. The name is probably not the best, but I started by dealing with libudev so that's what I named it. It probably would make sense to rename it if/when this gets merged. Eventually hopefully this can be turned on by default, but I'm not sure how this project would want to deal with setting the defaults.

I see there were a couple of attempts at doing some generic-ish backend work and bypassing hidapi. I looked at a branch or two, but ended up building the interaction in a way that felt like it made sense to me. Hopefully it's sensible for others too.

The `Box<dyn ...>` does mean an extra allocation but I think the alternative of `struct HidDevice<T: HidDeviceBackend>` would break API. An alternative would also be to embed the backend `HidDevice` and change the type of the field depending on the backend at compile time. That wouldn't really use the trait to make sure you've implemented everything though, but that's not the end of the world as it's all internal types anyway (and there are workarounds you can do in tests to enforce that, I think).

I added a couple of test files from a wireless mouse that I happen to have nearby to test that the usage and usage_page values are the same as hidapi.


src/linux_udev.rs
```diff
@@ -0,0 +1,762 @@
...
+    errno::Errno,
+    ioctl_read, ioctl_readwrite_buf,
+    poll::{poll, PollFd, PollFlags},
+    unistd::{read, write},
+};
+
+use super::{BusType, DeviceInfo, HidDeviceBackend, HidError, HidResult, WcharString};
+
+/// Global error to simulate what C hidapi does
+static GLOBAL_ERROR: Mutex<RefCell<Option<String>>> = Mutex::new(RefCell::new(None));
```
ruabmbua: You really do not need to emulate the C API, it would probably be a better idea to fix the API and make a breaking change. Why make the effort of implementing this in rust, and then keep bad C API practices ;-)

I imagine it would be pretty simple to remove check_error() API's when the udev backend is enabled.  For any actual legitimate use of this API, better to return the errors directly in all functions via Result.

src/linux_udev.rs
```diff
@@ -0,0 +1,762 @@
+//! Functions for talking to udev directly instead of going through hidapi
```
ruabmbua: Probably not a good name for the backend.

I would just call it linux.rs, or maybe linux_native.rs?

src/linux_udev.rs
```diff
@@ -0,0 +1,762 @@
...
+        };
+
+        Err(HidError::HidApiError {
+            message: msg.to_string(),
+        })
+    }
+}
+
+// Bus values from linux/input.h
+const BUS_USB: u16 = 0x03;
```
ruabmbua: But consts at the top of the file.

src/linux_udev.rs
```diff
@@ -0,0 +1,762 @@
...
+    if elems.len() != 3 || !elems.iter().all(Result::is_ok) {
+        return None;
+    };
+
+    let numbers: Vec<u16> = elems.into_iter().map(|n| n.unwrap()).collect();
+    Some((numbers[0], numbers[1], numbers[2]))
+}
+
+// From linux/hidraw.h
+const HIDRAW_IOC_MAGIC: u8 = b'H';
```
ruabmbua: put consts at top of file.

Even better, maybe put these ioctl defs and related consts into a external file.

src/linux_udev.rs
```diff
@@ -0,0 +1,762 @@
...
+        Err(_) => panic!("udev strings should always be utf8"),
+    }
+}
+
+/// Parse a HID_ID string to find the bus type, the vendor and product id
+///
+/// These strings would be of the format
+///     type vendor   product
+///     0003:000005AC:00008242
+fn parse_hid_vid_pid(s: &str) -> Option<(u16, u16, u16)> {
```
ruabmbua: Probably a good idea to out parsing related functionality (which might even be shared between backends) into a different file.
Also the unit tests for parsing could be moved to that file.

Cargo.toml
```diff
@@ -23,13 +23,18 @@ linux-static-libusb = []
 linux-static-hidraw = []
 linux-shared-libusb = []
 linux-shared-hidraw = []
+linux-shared-udev = ["dep:udev", "dep:nix"]
```
ruabmbua: hm, probably not a good name for the backend. Maybe rename it to just linux, and make it clear in the lib.rs documentation that the *linux* backend is a native rust implementation (mostly) without C dependencies.

ruabmbua (commented):
Just browsed through it at a quick glance, and found a few things (mostly code structure).
I will do a more in depth review when I have more time.

But thank you very much, getting C dependencies out of the way and then refactoring the API a bit would make the library much simpler to use (less dependency issues with multiple hidapi versions in the dep tree).

src/lib.rs
```diff
@@ -41,6 +41,7 @@
 //! - `linux-static-hidraw`: uses statically linked `hidraw` backend on Linux (default)
 //! - `linux-shared-libusb`: uses dynamically linked `libusb` backend on Linux
 //! - `linux-shared-hidraw`: uses dynamically linked `hidraw` backend on Linux
+//! - `linux-shared-udev`: uses dynamically linked `udev` backend on Linux
```
ruabmbua: udev is not really a backend for this, I`d rather call the backend *linux* and describe it as having no dependencies to the hidapi C library.

src/linux_udev.rs
carlosmn: I agree, I was wondering if it might be too forward to propose a PR to deprecate it, since it doesn't seem to return anything that you wouldn't have in a better way.

We could deprecate it in mainline and make it a no-op or just not even have it with non-C-hidapi backends if and when they come in. Mostly I added it because one of the examples uses it, so it would actually compile and run. If it goes away selectively then probably that would have to change.

carlosmn: Well it looks like this wouldn't be the first method that disappears depending on the backend, so maybe it's fine to just not expose it.

carlosmn:
I've renamed the backend and options to simply `linux` as ideally that would be the only/default one there.

carlosmn:
It looks like naming an option `linux` makes clippy think we're confused about what we mean, so now it's named `linux-native` and `linux_native`. This is probably for the best as it makes space in the naming for things like `macos_native`.

I've also edited some earlier commits so we open the right file so things do work.

Unfortunately by having the path be the dev node, i.e. `/dev/hidrawN`, the udev library doesn't seem to have a way to open based off of that, as it only has `Device::from_syspath` as a way to create the device directly, so we don't actually have enough information, so `get_device_info()` is currently simply broken.

src/linux_udev.rs
carlosmn: I've left the things here because I don't know if a macos_native backend would even need this or if the system libraries would parse a lot more for us.

src/linux_udev.rs
carlosmn: Sure, but nothing other than linux-native would use ioctls, no?

carlosmn:
Oh I also made it so the `check_error` methods are only available when using C hidapi.

src/linux_udev.rs
ruabmbua: yep these are pretty linux specific

ruabmbua:
Regarding udev problem with the devnode, can you maybe elaborate more? I think I had no issues with that when I was working on this myself. It is also poissible to map from /dev file to /sys path with udev (although I do not remember the exact API for that).

carlosmn:
In order to match the C hidapi does, and so you can pass the `DeviceInfo::path` into `HidApi::open_path` and have it work then we need to open via the dev node, and that's what we can store in our struct. I don't know off hand if the files under `/sys` can be opened to use as devices.

The way C hidapi then can get the info is via `udev_device_new_from_devnum` and then filling is as usual, but that constructor is not available in the Rust udev library. I don't see a way to go from one to the other without going through and listing them all, but I'm not that familiar with the layouts here.

carlosmn:
It looks like it should be easy yeah https://github.com/systemd/systemd/blob/5015b5014bcff93371aef2c78b92efcfc2e38a40/src/libsystemd/sd-device/sd-device.c#L304 though it would be better if the udev library supported it.

src/linux_udev.rs
carlosmn: Moved these to their own file that we then use in linux_native.

src/linux_udev.rs
carlosmn: Looking at what C hidapi does for macOS, it doesn't look like we'd need them for anything else. The vendor and product ids are parsed by the system and given to us, as are the usage_page and usage information. Same goes with Windows.

So we can move them somewhere else if you prefer, like with the ioctls, but it doesn't look like it's something we will be re-using.

carlosmn:
It turns out this completely broke macOS (and probably Windows) because they want to add methods so I guess I'll see if I can tackle that.

carlosmn:
It's not super clean but I've made it so each os can implements its own extras.

build.rs
```diff
@@ -77,19 +78,29 @@ fn compile_linux() {
...
                 pkg_config::probe_library("hidapi-libusb").expect("Unable to find hidapi-libusb");
                 println!("cargo:rustc-cfg=libusb");
+                println!("cargo:rustc-cfg=hidapi");
+            }),
+        ),
+        (
+            "LINUX_NATIVE",
+            Box::new(|| {
+                // The udev crate takes care of finding its library
+                println!("cargo:rustc-cfg=linux_native");
```
ruabmbua: Since linux_native is already a feature, this is not necessary. You can use the usual cfg(feature) attribute

src/error.rs
```diff
@@ -50,3 +50,20 @@ impl Display for HidError {
 }
 
 impl Error for HidError {}
+
+impl From<std::io::Error> for HidError {
+    fn from(e: std::io::Error) -> Self {
+        Self::HidApiError {
```
ruabmbua: You can a new io error enum variant, and use that instead of turning the error into a string. Allows the user to work with errors (e.g. for async extract EWOULDBLOCK / EAGAIN etc...).

src/error.rs
```diff
@@ -50,3 +50,20 @@ impl Display for HidError {
...
+            message: format!("{e}"),
+        }
+    }
+}
+
+#[cfg(linux_native)]
+impl From<nix::errno::Errno> for HidError {
+    fn from(e: nix::errno::Errno) -> Self {
```
ruabmbua, line 80: Probably also a better idea to properly store the error in the enum

src/lib.rs
```diff
@@ -57,21 +58,32 @@
...
+use winapi::shared::guiddef::GUID;
 
 mod error;
 mod ffi;
 
+#[cfg(hidapi)]
+mod hidapi;
+#[cfg(linux_native)]
+mod ioctl;
+#[cfg(linux_native)]
```
ruabmbua: probably better to not use manual cfg when not necessary. Can use cfg(feature)

src/lib.rs
```diff
@@ -287,74 +250,16 @@ impl HidApi {
...
-        interface_number: (*src).interface_number,
-        bus_type: (*src).bus_type,
-    })
-}
-
-#[derive(Clone)]
+#[derive(Clone, PartialEq)]
 enum WcharString {
     String(String),
+    #[cfg_attr(linux_native, allow(dead_code))]
```
ruabmbua: You could also just remove this enum variant in case of linux_native?

src/lib.rs
```diff
@@ -514,55 +419,85 @@ impl fmt::Debug for DeviceInfo {
...
+#[cfg(target_os = "macos")]
+trait HidDeviceBackend: HidDeviceBackendBase + HidDeviceBackendMacos {}
+#[cfg(target_os = "windows")]
+trait HidDeviceBackend: HidDeviceBackendBase + HidDeviceBackendWindows {}
+
+/// Automatically implement the top trait
+#[cfg(not(any(target_os = "macos", target_os = "windows")))]
+impl<T> HidDeviceBackend for T where T: HidDeviceBackendBase {}
+
+/// Automatically implement the top trait
```
ruabmbua, line 472: either only use this comment one time, or all three times

src/linux_native.rs
```diff
@@ -0,0 +1,670 @@
...
+const BUS_USB: u16 = 0x03;
+const BUS_BLUETOOTH: u16 = 0x05;
+const BUS_I2C: u16 = 0x18;
+const BUS_SPI: u16 = 0x1C;
+
+pub struct HidApiBackend;
+
+impl HidApiBackend {
+    pub fn get_hid_device_info_vector() -> HidResult<Vec<DeviceInfo>> {
+        // The C version assumes these can't fail, and they should only fail in case
```
ruabmbua, line 37: currently udev might only fail on memory alloc failure (which is hard to believe?).
You can also forward any errors happening in udev to the Result, and embed the udev error inside the enum.

src/linux_native.rs
```diff
@@ -0,0 +1,670 @@
...
+            Ok(e) => e,
+            Err(_) => return Ok(Vec::new()),
+        };
+        enumerator.match_subsystem("hidraw").unwrap();
+        let scan = match enumerator.scan_devices() {
+            Ok(s) => s,
+            Err(_) => return Ok(Vec::new()),
+        };
+
+        let mut devices = Vec::new();
```
ruabmbua: Maybe prefer a .filter_map() + collect for this code.
In this case the iterator code can give the vec hints on how much to reserve. Iterator code is also more idomatic.

src/linux_native.rs
```diff
@@ -0,0 +1,670 @@
...
+    }
+}
+
+/// Parse a HID_ID string to find the bus type, the vendor and product id
+///
+/// These strings would be of the format
+///     type vendor   product
+///     0003:000005AC:00008242
+fn parse_hid_vid_pid(s: &str) -> Option<(u16, u16, u16)> {
+    let elems: Vec<Result<u16, _>> = s.split(':').map(|s| u16::from_str_radix(s, 16)).collect();
```
ruabmbua: Do not create  a vector from the iterator. Instead store the iterator in a variable  and call .next() three times filling up the tuple.

src/linux_native.rs
```diff
@@ -0,0 +1,670 @@
...
+        None => return None,
+    };
+    let serial = match device.property_value("HID_UNIQ") {
+        Some(serial) => serial,
+        None => return None,
+    };
+    let path = match raw_device
+        .devnode()
+        .map(|p| p.as_os_str().to_os_string().into_vec())
+        .map(CString::new)
```
ruabmbua, line 106: I am not too happy that the lib is using CStrings for paths. Should be OsString or rather PathBufs.

Unfortunately getting rid of that will not be that easy. Something to consider when  the library is fully converted to native backends.

src/linux_native.rs
```diff
@@ -0,0 +1,670 @@
...
+        usage_page: 0,
+        usage: 0,
+        interface_number: -1,
+        bus_type,
+    };
+
+    // USB has a bunch more information but everything else gets the same empty
+    // manufacturer and the product we read from the property above.
+    let info = match bus_type {
+        BusType::Usb => fill_in_usb(raw_device, info, name),
```
ruabmbua, line 130: Probably clearer to pass the info as &mut reference and not use a input + return parameter?

src/linux_native.rs
```diff
@@ -0,0 +1,670 @@
...
+        release_number,
+        manufacturer_string,
+        product_string,
+        interface_number,
+        ..info
+    }
+}
+
+// From linux/hidraw.h
+struct HidrawReportDescriptor {
```
ruabmbua: Why not simply store this in a vector or boxed slice?
You can still newtype HidrawReportDescriptor around it.

ruabmbua: Also using u32 for a length in rust is weird. please use usize

src/linux_native.rs
```diff
@@ -0,0 +1,670 @@
...
+    ///
+    /// It returns an error if the value slice is too large for it to be a HID
+    /// descriptor
+    #[cfg_attr(not(test), allow(dead_code))]
+    pub fn from_slice(value: &[u8]) -> HidResult<Self> {
+        let size: u32 = match value.len().try_into() {
+            Ok(v) => v,
+            Err(_) => {
+                return Err(HidError::HidApiError {
+                    message: "HID report descriptor over 4kB".into(),
```
ruabmbua: Not sure why there is a 4kb error message here. The try_into::<u32>() will be able  to convert much larger values.

ruabmbua: yep this panics: 

```rust
    #[test]
    fn overflow() {
        let data = vec![0u8; 10 * 4096];
        let desc = HidrawReportDescriptor::from_slice(&data).expect("bla");
    }
}
```

src/linux_native.rs
```diff
@@ -0,0 +1,670 @@
...
+    fn next(&mut self) -> Option<Self::Item> {
+        let (usage_page, page, advanced) =
+            match next_hid_usage(self.value, self.initial, self.usage_page) {
+                Some(n) => n,
+                None => return None,
+            };
+
+        self.usage_page = usage_page;
+        self.initial = false;
+        self.value = &self.value[advanced..];
```
ruabmbua: probably makes more sense to store a offset instead of initial bool, and not resize the slice in the iterator, but rather track with offset?

Then initial property can be found by checking if offset == 0

ruabmbua: Also this whole thing looks like it could use a `Cursor`, which would keep track of bytes read. Works like a file with .read() .seek() etc...

src/linux_native.rs
```diff
@@ -0,0 +1,670 @@
...
+
+/// Get the bytes from a HID report descriptor
+///
+/// Must only be called with `num_bytes` 0, 1, 2 or 4.
+fn hid_report_bytes(desc: &[u8], num_bytes: usize) -> u32 {
+    if num_bytes >= desc.len() {
+        return 0;
+    }
+
+    match num_bytes {
```
ruabmbua: Probably makes more sense to not match but rather just do  this:

```rust
let mut bytes: [0u8; 4] = [0; 4];
bytes[..num_bytes].copy_from_slice(desc[..num_bytes]);
u32::from_le_bytes(bytes)
```

Also the num_bytes param may be unneeded? Why not just use the slice len

src/linux_native.rs
```diff
@@ -0,0 +1,670 @@
...
+            return Some((desc[1].into(), 3));
+        }
+
+        // Malformed report
+        return None;
+    }
+
+    // Short Item. Bottom two bits contains the size code
+    match key & 0x03 {
+        v @ 0 | v @ 1 | v @ 2 => Some((v.into(), 1)),
```
ruabmbua: v @ 0..=2 ?

src/linux_native.rs
```diff
@@ -0,0 +1,670 @@
...
+    dev.attribute_value(attr)
+        .map(Into::into)
+        .map(osstring_to_string)
+        .unwrap_or(WcharString::None)
+}
+
+/// Get the attribute from the device and convert it into a i32
+///
+/// On error or if the attribute is not found, it returns -1;
+fn attribute_as_i32(dev: &udev::Device, attr: &str) -> i32 {
```
ruabmbua: I would rather have API`s using Option / Result instead of negative error codes.

src/linux_native.rs
```diff
@@ -0,0 +1,670 @@
...
+    dev.attribute_value(attr)
+        .and_then(OsStr::to_str)
+        .and_then(|v| i32::from_str_radix(v, 16).ok())
+        .unwrap_or(-1)
+}
+
+/// Get the attribute from the device and convert it into a u16
+///
+/// On error or if the attribute is not found, it returns 0;
+fn attribute_as_u16(dev: &udev::Device, attr: &str) -> u16 {
```
ruabmbua: yep, please use options intead. Its hard to keep  track of all those error == 0, error < 0, error == -1 etc...
Not necessary with rust

src/linux_native.rs
```diff
@@ -0,0 +1,670 @@
...
+        .and_then(OsStr::to_str)
+        .and_then(|v| u16::from_str_radix(v, 16).ok())
+        .unwrap_or(0)
+}
+
+/// Convert a [`OsString`] into a [`WcharString`]
+fn osstring_to_string(s: OsString) -> WcharString {
+    match s.into_string() {
+        Ok(s) => WcharString::String(s),
+        Err(_) => panic!("udev strings should always be utf8"),
```
ruabmbua, line 389: are you sure? This is probably more related to the locale of the linux system, which does not force utf-8

src/linux_native.rs
```diff
@@ -0,0 +1,670 @@
...
+    };
+
+    let numbers: Vec<u16> = elems.into_iter().map(|n| n.unwrap()).collect();
+    Some((numbers[0], numbers[1], numbers[2]))
+}
+
+/// Object for accessing the HID device
+pub struct HidDevice {
+    blocking: Cell<bool>,
+    fd: OwnedFd,
```
ruabmbua, line 425: Oh nice new struct from new rust, did not know about it

src/linux_native.rs
```diff
@@ -0,0 +1,670 @@
...
+        })
+    }
+
+    pub(crate) fn open_path(device_path: &CStr) -> HidResult<HidDevice> {
+        // Paths on Linux can be anything but devnode paths are going to be ASCII
+        let path = device_path.to_str().expect("path must be utf-8");
+        let fd: OwnedFd = match OpenOptions::new()
+            .read(true)
+            .write(true)
+            .custom_flags(libc::O_CLOEXEC | libc::O_NONBLOCK)
```
ruabmbua, line 458: I am not sure CLOEXEC is useful to pass as a default?

src/ioctl.rs
```diff
@@ -0,0 +1,29 @@
...
+const HIDRAW_IOC_MAGIC: u8 = b'H';
+const HIDRAW_IOC_GRDESCSIZE: u8 = 0x01;
+const HIDRAW_SET_FEATURE: u8 = 0x06;
+const HIDRAW_GET_FEATURE: u8 = 0x07;
+
+ioctl_read!(
+    hidraw_ioc_grdescsize,
+    HIDRAW_IOC_MAGIC,
+    HIDRAW_IOC_GRDESCSIZE,
+    i32
```
ruabmbua: probably should use the c_int type instead to make sure this matches with the kernel on all weird archs?

src/linux_native.rs
```diff
@@ -0,0 +1,670 @@
...
+    }
+}
+
+impl HidDeviceBackendBase for HidDevice {
+    fn write(&self, data: &[u8]) -> HidResult<usize> {
+        if data.is_empty() {
+            return Err(HidError::InvalidZeroSizeData);
+        }
+
+        Ok(write(self.fd.as_raw_fd(), data)?)
```
ruabmbua, line 500: This might return EAGAIN / EWOULDBLOCK, even if the device should be running in blocking mode?

ruabmbua (changes requested):
I did  a more in depth review, hope its not too much ;-)

carlosmn:
Heh, it looks a bit too much for the PR view :) I'll go through the comments. A lot looks like the result of porting the C code to Rust and keeping some of its behaviour.

build.rs
carlosmn: It's following the pattern elsewhere of adding the extra cfg but yeah I think it makes sense to skip it.

src/lib.rs
carlosmn: I was trying to avoid changing API but I now realise this is just internal to the library so that's maybe a slightly more efficient option.

src/linux_native.rs
carlosmn: It would require a version bump due to breaking the API so yeah i think it makes sense to only do it once everything can change over.

src/linux_native.rs
carlosmn: It seemed more rusty for the function to take ownership and return the changed one, kinda like you do with a builder pattern. But updating the fields also works.

src/linux_native.rs
carlosmn: It's just what the C version uses, which is not what we would do fresh in Rust. We probably could thinly-wrap a `Vec<u8>` or something.

src/linux_native.rs
carlosmn: 4kB is the max size of a HID report descriptor from sysfs, at last according to C hidapi. If something makes us read a file that's bigger, that's  probably a bug somewhere. That said, I think I missed out on an `if ` in the match or something because that error message definitely doesn't match the actual error.

src/linux_native.rs
carlosmn: I think that does work. It's clearer now that the endian conversion isn't done by hand. The `num_bytes` is the actual amount of data that we should read, while `desc` is the whole rest of the descriptor. Now, we should be able able to call this function with the equivalent of `&desc[..num_bytes]` in the parent, but I'm not sure where it's less awkward.

src/linux_native.rs
carlosmn: Yeah, that'd make sense as a rustification pass.

src/linux_native.rs
carlosmn: I can double-check but I think this is mostly about properties in the HID stuff that don't change (at least the ones we use).

src/linux_native.rs
carlosmn: The worry here would be to increase the minimum version of Rust too much as 1.63, where this was stabilised, is less than a year old.

src/ioctl.rs
carlosmn: I don't know if there are arches where Linux's int isn't 32-bit but it doesn't hurt.

src/linux_native.rs
carlosmn: Good question, this is what the C hidapi does so it might be fine? This isn't a regular file write but this ends up enqueuing into whatever queue the kernel has so I don't know if it this could block.

src/linux_native.rs
carlosmn: It's a useful default in general. Without it, a child process would also receive a handle to any hid devices the parent process has open, which seems like a bad idea.

ruabmbua: You are right, probably the better default. I use it everywhere in c++ also

src/lib.rs
carlosmn: Ok I just tried this, but if we remove `Raw` on linux-native, then all the other code that wants to use it in `DeviceInfo` breaks and that is public API so it seems better to leave it at allowing dead code.

src/lib.rs
carlosmn: Ah, this came from doing it in a VM where copy-paste was annoying so I missed out on the comment.

src/linux_native.rs
carlosmn: Based on https://docs.kernel.org/hid/hidraw.html where it only mentions blocking on `read()` it sounds like this is more akin to putting stuff in a queue than dong IO.

src/linux_native.rs
carlosmn: This went away with the other changes around the cursor.

src/linux_native.rs
carlosmn: I've ported it to use `Cursor` which is nicer ins some ways but in a bunch of places we really do just want to grab a single byte so we end up doing `cursor.bytes().next()` a couple of times, and the size to advance is also based off of the initial position, so a couple of places are a bit awkward, but it does remove a bit of state from the iterator itself.

src/linux_native.rs
carlosmn: This is now just wrapping a vector, so you could have descriptors larger than the kernel allows, but that's probably fine since the kernel is giving us the descriptor.

src/linux_native.rs
carlosmn: We return a vec so we still need a flatten after this. I suppose we could return an empty vector instead of a None and then a `flatten_map` by itself would work, but the semantics are slightly different then.

src/linux_native.rs
carlosmn: With all the enums it gets awkward but I made a macro to behave like let-else so it's a bit nicer to read and we do avoid allocating the two vectors.

src/error.rs
carlosmn: I didn't want to add to the public API here and break error handling that users of the library might already have, but for this it's probably OK since making use of more rust probably means this is going to be a fairly normal error that we'll have to deal with.

src/error.rs
carlosmn: This feels like a different situation from `std::io::Error` as it would add a variant that only exists with some options, which seems like it's going to make it a pain for users of the library to deal with. But if `std::io::Error` does become a variant, then maybe we can make a decent effort at translating it to that.

src/linux_native.rs
carlosmn: Right, on Linux you would basically never see an actual malloc error, so we could just blow up there. I'm not a fan of adding udev error structs into the library's errors for the same reason as nix errors, where they exist just on the one OS.

src/linux_native.rs
```diff
@@ -0,0 +1,640 @@
...
+    }
+}
+
+/// Get the bytes from a HID report descriptor
+///
+/// Must only be called with `num_bytes` 0, 1, 2 or 4.
+fn hid_report_bytes(cursor: &mut Cursor<&Vec<u8>>, num_bytes: usize) -> u32 {
+    let mut bytes: [u8; 4] = [0; 4];
+    if cursor.read_exact(&mut bytes[..num_bytes]).is_err() {
+        return 0;
```
ruabmbua: This should probably also be returned as an Err/option. And on the call site treat it as a parsing error discarding the whole item?

src/error.rs
carlosmn: I've added a variant for this as it's generally useful and we'll need it more as we write more Rust.

src/error.rs
carlosmn: Given that any nix Errno is going to be convertible into a `std::io::Errror` and they provide a impl for that, I've made it so nix errors convert into the `IoError` variant which lets us avoid an extra equivalent variant.

src/linux_native.rs
carlosmn: Yes I suppose that's what you should be doing in Rust.

src/linux_native.rs
ruabmbua: Ok sure its ok

src/linux_native.rs
ruabmbua: yes lets keep it as is

src/linux_native.rs
ruabmbua: Ok interesting. Although its always a bit weird when special devices in linux do not behave like e.g. the read() or write() manpages describe it.

But what happens if the queue is full? Maybe it will block then. Could be a theoretical possibility, that has no chance of practically happening. Lets keep it like it is, if we ever have weird issues in regards to writing, we can look again.

src/error.rs
ruabmbua: Ok thx

src/error.rs
ruabmbua: Good idea, lets do it like that

src/linux_native.rs
carlosmn: Looking through a bit of kernel code, it looks like the `hidraw_write` function does a bunch of checking and you just end up doing the same as the send-report ioctl, and then it gets handed over to the appropriate driver to whatever callback it has decided. So it looks like what happens is whatever each device driver decides to do.

src/linux_native.rs
```diff
@@ -0,0 +1,644 @@
...
+        }
+    };
+}
+
+/// Parse a HID_ID string to find the bus type, the vendor and product id
+///
+/// These strings would be of the format
+///     type vendor   product
+///     0003:000005AC:00008242
+fn parse_hid_vid_pid(s: &str) -> Option<(u16, u16, u16)> {
```
ruabmbua, line 413: This also works:

```rust
fn parse_hid_vid_pid(s: &str) -> Option<(u16, u16, u16)> {
    let mut elems = s.split(':').map(|s| u16::from_str_radix(s, 16));
    let devtype = elems.next()?.ok()?;
    let vendor = elems.next()?.ok()?;
    let product = elems.next()?.ok()?;

    Some((devtype, vendor, product))
}
```

carlosmn: oh huh, is this a newer feature? I thought I'd tried to do this at some point in the past but `Option` wasn't a thing that the question mark worked with.

src/linux_native.rs
carlosmn: We do use this for the product and manufacturer, maybe those don't have to be utf-8 then. I wonder what to do here, since we don't have the wide chars for the `WcharString::Raw`. Should we do a lossy conversion? Convert to wchar if it's not utf-8?

src/linux_native.rs
ruabmbua: Not sure when, but already a while I think.

ruabmbua:
Thanks for the work, I will merge this back now and do some small fixups. I will then release it to crates.io after I have some more time verifying it does not break anything (hopefully I have time soon).

carlosmn:
Awesome, glad we could get to something we're happy with.

PR #108, opened 2023-04-15
Make the WcharString to Option<String> conversion into a From

carlosmn:
Clippy is being annoying about this and it's generally good practice to implement the `From` which then automatically gets us the `Into`.


ruabmbua:
Good find, thx for the fix.

PR #118, opened 2023-05-20
Implement Debug for HidDevice

marienz:
HidDevice previously implemented Debug (see
ed7bc23ea9d47e1f98e9394e3c478580a68df215), but the copy of the public API added in 7439af69b91164a70b3d1b54a6f5c6d54ebe7749 did not. This restores that functionality.

This just prints `HidDevice { .. }`, but that is still useful as it makes it easy for wrapping types to implement Debug (it makes `derive(Debug)` on structs containing a HidDevice work).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -482,6 +482,12 @@ pub struct HidDevice {
     inner: Box<dyn HidDeviceBackend>,
 }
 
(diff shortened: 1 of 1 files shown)


ruabmbua:
Thx, that was an oversight when we refactored some of the code.

PR #125, opened 2023-07-08
Add support for get_report_descriptor()

### LINKED ISSUE #104: Add `get_report_descriptor`
`hidapi` recently added HID report descriptor parsing in https://github.com/libusb/hidapi/pull/451 for the (upcoming?) 0.14.0 version. Can it be made accessible here?

RavuAlHemio:
Fixes #104.

Successfully tested on Windows, macOS and with the following Linux backends:

* linux-shared-hidraw
* linux-static-hidraw
* linux-native

Fails with the following Linux backends:

* linux-shared-libusb
* linux-static-libusb

(Both return `hid_error is not implemented yet` when calling `HidApi::open`.)

I haven't tested it on Illumos.

### DIFF
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -86,6 +86,11 @@ extern "C" {
         string: *mut wchar_t,
         maxlen: size_t,
     ) -> c_int;
+    pub fn hid_get_report_descriptor(
+        hid_device: *mut HidDevice,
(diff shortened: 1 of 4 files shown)


ruabmbua:
Thanks for picking this up and making it work with linux-native backend!

PR #126, opened 2023-07-08
Extend CI/CD to all Linux combinations + macOS + Windows

RavuAlHemio:
Sure, it's not a test suite (kinda hard to do that without some emulation of HID hardware), but at least it should catch build issues.


ruabmbua:
Thanks!

Issue #127, opened 2023-07-14
Assertion / Illegal instruction on macOS when calling `HidApi::new` in different threads


Tiwalun:
I've run into an issue on macOS where calling `HidApi::new` from different threads leads to an abort of the process, with an exception for an illegal instruction. 

Looking at the stacktrace, it seems to be cause by a `CFAssertMismatchedTypeID` assertion somewhere in Core Foundation.

Are there any special precautions that have to be taken when using `hidapi` from multiple threads? I couldn't find anything in the documentation.

## Stacktrace

```
* thread #5, name = 'try 1', stop reason = EXC_BAD_INSTRUCTION (code=EXC_I386_INVOP, subcode=0x0)
  * frame #0: 0x00007ff80cf83f3e CoreFoundation`_CFAssertMismatchedTypeID + 110
    frame #1: 0x00007ff80ce254bd CoreFoundation`CFRunLoopAddSource + 973
    frame #2: 0x00007ff80f76e305 IOKit`IOHIDDeviceScheduleWithRunLoop + 74
    frame #3: 0x00007ff80f772651 IOKit`__IOHIDManagerDeviceApplier + 527
    frame #4: 0x00007ff80f735537 IOKit`__IOHIDManagerDeviceAdded + 766
    frame #5: 0x00007ff80f73513b IOKit`__IOHIDManagerSetDeviceMatching + 347
    frame #6: 0x0000000100010724 hidapi-repro`hid_enumerate(vendor_id=0, product_id=0) at hid.c:654:2
    frame #7: 0x000000010000ce03 hidapi-repro`hidapi::hidapi::HidApiBackend::get_hid_device_info_vector::hdb756fded0153d2d at hidapi.rs:25:36
    frame #8: 0x000000010000dbde hidapi-repro`hidapi::HidApi::new::h471133b553ec121f at lib.rs:171:27
    frame #9: 0x0000000100006141 hidapi-repro`hidapi_repro::enumerate::h2a623970c0b3f1ae at main.rs:16:13
```

Repo with code to reproduce: https://github.com/tiwalun/hidapi-repro

ruabmbua:
Hi! Because of the not very well defined nature of the underlying C library (regarding multithreading), we used to prevent creating a second instance of a Hidapi object, by acquiring a lock.
However people started requesting that I lift this restriction, and testing on the backends that we could test showed, that as long as you open a device not from multiple threads, everything works out fine.

[...]

PR #131, opened 2023-07-23
feat: Native macOS backend

Tiwalun:
This is a port of the hidapi code for macOS to pure Rust, with some small modifications to make it more thread safe.

I've tested it with [probe-rs](https://github.com/probe-rs/probe-rs), and it seems to work fine so far.

The error handling could be improved, but that might be better to do in coordination with the other native backends.

## Remaining tasks

- [ ] Implement open options (exclusive / non-exclusive)
- [ ] Handle blocking / non-blocking reads
- [ ] Check thread-safety


ruabmbua:
Hi, thank you for your work! I will hopefully have some time to review it in the coming week.

src/macos_native.rs
```diff
@@ -0,0 +1,892 @@
...
+
+#[derive(Debug)]
+struct WrappedCFRunLoop(CFRunLoop);
+
+// This should be safe, according to documetation:
+// https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Multithreading/RunLoopManagement/RunLoopManagement.html#//apple_ref/doc/uid/10000057i-CH16-SW26>
+//
+// It is recommended that changing the configuration a run loop should be done from the thread that owns the run loop whenever possible.
+// In the use here, we only use the wrapper to wake up the run loop, which is safe.
+unsafe impl Send for WrappedCFRunLoop {}
```
micolous, line 91: This should no longer be needed as of https://github.com/servo/core-foundation-rs/pull/610

src/macos_native.rs
```diff
@@ -0,0 +1,892 @@
...
+
+impl HidApiBackend {
+    pub fn get_hid_device_info_vector() -> HidResult<Vec<DeviceInfo>> {
+        HidApiBackend::enumerate(None, None)
+    }
+
+    fn enumerate(vendor_id: Option<u16>, product_id: Option<u16>) -> HidResult<Vec<DeviceInfo>> {
+        let manager = IOHIDManager::create();
+
+        if vendor_id.is_some() || product_id.is_some() {
```
micolous, line 120: Allow matching by Usage/Usage Page, and multiple VIDs/PIDs.

Without any matching rules, macOS demands "Input Monitoring" permission, which is quite bad.

src/macos_native.rs
```diff
@@ -0,0 +1,892 @@
...
+
+    device: Mutex<IOHIDDevice>,
+
+    disconnected: AtomicBool,
+    shutdown_thread: AtomicBool,
+    shutdown_barrier: Barrier,
+
+    // Condition variable linked to input_reports
+    condition: std::sync::Condvar,
+    input_reports: Mutex<VecDeque<Vec<u8>>>,
```
micolous, line 80: Use an `mpsc::channel` rather than a `VecDeque`. That way you get queue size limits for free, you don't need to wrap everything in a `Mutex`.

You also get helpers like `recv_timeout`...

src/macos_native.rs
```diff
@@ -0,0 +1,892 @@
...
+    // Notify that the thread is stopping
+    {
+        let _guard = shared_state.input_reports.lock().unwrap();
+        shared_state.condition.notify_all();
+    }
+
+    {
+        let device = shared_state.device.lock().unwrap();
+
+        // unregister the input report callback
```
micolous, line 832: This isn't necessary - `IOHIDDevice` automatically unregisters these on close: https://github.com/apple-oss-distributions/IOKitUser/blob/b0b3f822b7507c265aa8a1e37c3100c03ca82039/hid.subproj/IOHIDDevice.c#L483-L486

src/macos_native.rs
```diff
@@ -0,0 +1,892 @@
...
+        *run_loop = Some(WrappedCFRunLoop(CFRunLoop::get_current()));
+    }
+
+    barrier.wait();
+
+    while (!shared_state.shutdown_thread.load(Ordering::Relaxed))
+        && (!shared_state.disconnected.load(Ordering::Relaxed))
+    {
+        // TODO: Verify timeout value
+        let code = unsafe { CFRunLoopRunInMode(run_loop_mode.as_concrete_TypeRef(), 1000.0, 0) };
```
micolous, line 806: I'm struggling to understand the design here, but why not just `CFRunLoopRun()`? It does basically the same thing as your loop: https://github.com/apple-oss-distributions/CF/blob/dc54c6bb1c1e5e0b9486c1d26dd5bef110b20bf3/CFRunLoop.c#L2676-L2682

On disconnect/`Drop`, you can issue a `CFRunLoopStop()` from the main thread – `HidDevice::drop` already has a reference to the device's `CFRunLoop`.

I don't think it matters in this loop whether the device was closed by the application (terminating the run loop) or there was an explicit disconnect (terminating the run loop); a disconnected device is a disconnected device. :)

`mpsc::channel` also saves you here too, because `Drop`ing the `Sender` will unblock any pending `Receiver::recv` calls on another thread, and make future calls fail.

If you've only associated a single `IOHIDDevice` with the `CFRunLoop`, the run loop will stop on disconnect _anyway_. `hid_removal_callback` appears to only set a flag that it was disconnected, but this could _also_ be detected by setting your own (not-mutex-guarded) flag on `close()`.

micolous (commented):
Having written something similar myself, this design is _really_ strange. Rust's standard library has functionality which could make the implementation _significantly_ simpler.

It looks like this attempts to decouple `IOHIDDevice`s from the `IOHIDManager` which created it, in order to allow "opening by path" (which is nonsensical on macOS). It would be better to attach that to the `HidApiBackend`.

While this may be more of a comment on the state of `hidapi`'s interface, the device matching rules available are _severely_ lacking (only single VID+PID), which is a problem in modern versions of macOS, which put free-for-all access behind an "input monitoring" permission. There's also no way to monitor device insertion/removal events, so you're forced to continually prod `HidApiBackend::enumerate()`.

src/macos_native/ffi.rs
```diff
@@ -0,0 +1,469 @@
...
+                matching
+                    .map(|m| m.as_concrete_TypeRef())
+                    .unwrap_or(std::ptr::null()),
+            );
+        }
+    }
+
+    pub fn copy_devices(&self) -> Vec<IOHIDDevice> {
+        let set: CFSet<IOHIDDeviceRef> = unsafe {
+            let set = IOHIDManagerCopyDevices(self.as_concrete_TypeRef());
```
micolous, line 89: This isn't documented, but `IOHIDManagerCopyDevices` can return `nullptr` if there are no matching devices.

src/macos_native.rs
```diff
@@ -0,0 +1,892 @@
...
+            return Err(HidError::HidApiError {
+                message: "failed to open IOHIDDevice from mach entry".into(),
+            });
+        }
+
+        let max_input_report_len = device
+            .get_i32_property("MaxInputReportSize")
+            .unwrap_or_default();
+
+        let run_loop_mode = format!("HIDAPI_{:p}", device.as_concrete_TypeRef());
```
micolous, line 600: This isn't right, but [the `IOHIDDeviceUnscheduleFromRunLoop` `runLoopMode` parameter is unused anyway](https://github.com/apple-oss-distributions/IOKitUser/blob/b0b3f822b7507c265aa8a1e37c3100c03ca82039/hid.subproj/IOHIDDevice.c#L749-L753).

src/macos_native.rs
micolous: Correction: still needed, the PR made it into v0.8.5, but it was yanked due to https://github.com/servo/core-foundation-rs/issues/619.

src/macos_native.rs
Tiwalun: Could you explain what you mean by not right? Using custom run modes seems to be possible according to the documentation at <https://developer.apple.com/documentation/corefoundation/cfrunloop?language=objc>.

Or do you mean that a custom run mode shouldn't be used?

src/macos_native.rs
Tiwalun: Right now, the oldest received input report is dropped when the limit is reached. I don't think this is possible with the `mpsc::sync_channel`, the size limit there could only be used stop sending new input reports, but it doesn't seem possible to remove elements from the channel.

Using a channel would be nice, but it would change the behaviour compared to the `C` hidapi library.

src/macos_native.rs
Tiwalun: This would require changes to the other backends as well, or additional Mac-specific functions.

Tiwalun:
Thanks for looking at this! It's good to get some feedback from somebody who knows these APIs.

> Having written something similar myself, this design is _really_ strange. Rust's standard library has functionality which could make the implementation _significantly_ simpler.

As mentioned in the description, this is more or less a direct port of the hidapi C library, with only minor changes to work around the problem mentioned in #127.

I'm not sure what the requirements regarding compatibilty with the existing implementation are. If this is only an optional backend, and the hidapi library is kept as an alternative, then it should probably be compatible, which limits the amount of changes that can be done. On the other hand, if compatibility is not a concern, then some of the improvements like using a `mpsc::channel` could be done more easily.

@ruabmbua What are your thoughts regarding compatibility with the C library?

ruabmbua:
Thanks for the review and the work done! I already started with it myself, but could not motivate myself to start learning macos API`s ^^.

Regarding compatibility to the current public API of the library: I would keep it as is for now (with non breaking changes allowed). This means I agree that the new optional pure rust macos backend should try to mimic the old one as close as it can.

I might revisit the public API in the future (v3.0.0), and do mostly some fixes that make it more idomatic rust.
I am not sure about async as of yet, I struggle to see the large benefit of making the library true async. Its not likely,
that an application will need to handle a huge amount of hid devices in comparison to e.g. network connections.
You can always spawn dedicated worker threads for devices, and integrate them with e.g. async channels + an API wrapper into e.g. tokio. Maybe I am missing a use case, where a true async implementation for hidapi would help.

carlosmn:
> I am not sure about async as of yet, I struggle to see the large benefit of making the library true async.

The main functionality I would want out of async in this library is being able to have async read and write from/to a device. Some might not be possible to actually do async and some might make more sense to use something else (like updating the list of relevant devices). But I don't see a way around async read/write without helf from the library.

> You can always spawn dedicated worker threads for devices, and integrate them with e.g. async channels + an API wrapper into e.g. tokio. Maybe I am missing a use case, where a true async implementation for hidapi would help.

Unfortunately you cannot always replicate async. The device I'm most interested in driving is interactive, so while I can (and do) have a thread that takes care of talking to the device, it needs to choose whether to read or write.

I want to react to the user's interaction quickly, so it's waiting on a read. But I have to be able to write to the device based on what might be happening on the computer. Right now the only way this works is by providing a short-but-not-too-short timeout to the read. If the timeout is too long, the interaction becomes too sluggish. If it's too short, we're looping and wasting cycles and power on reading nothing.

There's two ways of making this better that I can see: async or cancellable timeouts, and both would need some support from the library, and async might be the easier one overall.

The library does currently implement `AsFd` on the linux-native backend, which in principle would let you run `poll` or whatever yourself and add an extra eventfd or something, but that only works if you're using the library directly. What I'm doing is complex enough that there's another library in between, and it shouldn't have to know about the different tricks on the different OSs to do this, or expose a file descriptor. This would only work on Linux anyway and any of the other systems would need some other way of handling this.

Which is why ISTM that at least having async read and write would help quite a bit and it would allow users of this library not to have to implement workarounds for each platform. And the native Linux and Windows as well as a libusb implementation should all be able to do some form of asynchornous IO (IIRC what the C hidapi does is force the libusb event stuff to become blocking).

I've been playing around with what an async interface would look like but I haven't quite figured out what it should  look like. A different `HidDevice` struct, or specific async methods on it, something completely different? It's also awkward trying not to enforce a particular runtime.

davehorner:
what's the state of this @Tiwalun?  I can give this a try...

src/macos_native.rs
micolous: Ah, I was getting at that it was a custom mode, which I didn't realise was thing you could do. And I thought it was unused as I didn't trace where else that parameter was going.

In any case, [I think you could get away with using the default runloop mode](https://github.com/ruabmbua/hidapi-rs/pull/131/files#r1323951177). This code will create a new thread for each device connection, and there's a runloop for every thread, so there's no other events/observers/timers to isolate from.

src/macos_native.rs
micolous: > Using a channel would be nice, but it would change the behaviour compared to the `C` hidapi library.

I didn't realise [`hidapi` did this on macOS](https://github.com/libusb/hidapi/blob/f18d2c0768bed300d13758dd9f52b57163c0498f/mac/hid.c#L893-L914).

FWIW, [the Linux version of `hidapi` just reads from a file handle](https://github.com/libusb/hidapi/blob/f18d2c0768bed300d13758dd9f52b57163c0498f/linux/hid.c#L1179).

I've observed a USB _powered_ VGA display [^1] become unresponsive to OSD button presses if there was nothing requesting HID input reports, which was an issue when plugging it into a Linux machine (Windows was fine). That makes me think using Linux APIs would stall out a device if input reports weren't handled fast enough, and it depends on the device firmware as to how much it can stall by, and which events would be lost while stalling.

[Microsoft's docs say that HID input reports are stored in a ring-buffer](https://learn.microsoft.com/en-us/windows-hardware/drivers/hid/troubleshooting-hid-reports#dropped-hid-reports), but it's not specified whether the oldest or newest report is dropped when it is full.

[^1]: The manufacturer put an undocumented debugging interface on the USB interface. This was long before USB wall-socket chargers and USB battery packs were commonplace, so "only powering it from a computer" was a reasonable assumption, and I presume they only ever tested it on Windows PCs.

micolous (commented):
I missed this on my first review, but if the goal is "do a port of what `hidapi` does in C to Rust", then your strategy is a lot more understandable.

Unfortunately, you end up being limited to doing what C can do. 🤷

PR #135, opened 2023-11-10
Remove the need to panic when calling new() and new_without_enumerate()

JonathanLindsey:
This removes the need to panic when `HidApi::new()` is called in one place and `HidApi::new_without_enumerate()` is called in another place.

I am assuming it is ok to just rerun some of the initialization code when `new()` is called and `new_without_enumerate()` has been called previously. This probably could be optimized further on certain platforms.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -117,35 +117,41 @@ fn lazy_init(do_enumerate: bool) -> HidResult<()> {
 
     match *init_state {
         InitState::NotInit => {
-            #[cfg(all(libusb, not(target_os = "freebsd")))]
-            if !do_enumerate {
-                // Do not scan for devices in libusb_init()
-                // Must be set before calling it.
-                // This is needed on Android, where access to USB devices is limited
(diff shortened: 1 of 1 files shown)


ruabmbua:
I am not sure how useful this is. Also, this would require verifying that the C code does not break this invariant every time we update the underlying C library.

We use the HidApi::new also, to make sure its the only HidApi object alive in the program. This helps guarding the underlying C function from racing in multi threaded scenarios.

Eventually when we have good rust backends for all the major platforms, the plan is to start changing the API surface to be a bit more idiomatic. Making it usable from multiple threads, and removing global state is of course one of the things that would be nice.

JonathanLindsey:
The reason for this (or something like it) was so that if `new()` was called in one place in a program and `new_withoutout_enumerate()` was called in another place (possibly by different crates), it would still work. Now that I've looked at this, it seems harder than I though to get right.

Now I'm thinking this was buggy when upgrading from non-enumerating to enumerating mode or other things.

PR #138, opened 2024-01-10
Simplified conditional compilation

sidit77:
I recently fount out that this crate currently fails to compile on non-windows platforms when the `windows-native` feature is active.  This is caused by a missing `target_os = "windows"` in one of the `#[cfg(...)]` attributes.

I refactored the individual `#[cfg(...)]` attributes to use the `cfg_if` macro instead to improve readablillity and hopefully prevent such mistakes in the future.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -58,28 +58,8 @@
 //! [`HidDevice`] handles can access the same physical device. For backward compatibility this is
 //! an opt-in that can be enabled with the `macos-shared-device` feature flag.
 
-#[cfg(target_os = "windows")]
-use windows_sys::core::GUID;
-
 mod error;
 mod ffi;
-
-#[cfg(hidapi)]
-mod hidapi;
-
-#[cfg(all(feature = "linux-native", target_os = "linux"))]
-#[cfg_attr(docsrs, doc(cfg(all(feature = "linux-native", target_os = "linux"))))]
-mod linux_native;
-#[cfg(target_os = "macos")]
-#[cfg_attr(docsrs, doc(cfg(target_os = "macos")))]
-mod macos;
-#[cfg(target_os = "windows")]
-#[cfg_attr(docsrs, doc(cfg(target_os = "windows")))]
-mod windows;
-
-#[cfg(feature = "windows-native")]
-#[cfg_attr(docsrs, doc(cfg(all(feature = "windows-native", target_os = "windows"))))]
-mod windows_native;
 
 use libc::wchar_t;
 use std::ffi::CStr;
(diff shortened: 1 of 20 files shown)


src/windows_native/descriptor/encoder.rs
```diff
@@ -1,7 +1,7 @@
-use crate::windows_native::descriptor::typedefs::{Caps, LinkCollectionNode};
-use crate::windows_native::descriptor::types::{ItemNodeType, Items, MainItemNode, MainItems};
-use crate::windows_native::error::{WinError, WinResult};
-use crate::windows_native::utils::PeakIterExt;
+use super::typedefs::{Caps, LinkCollectionNode};
+use super::types::{ItemNodeType, Items, MainItemNode, MainItems};
+use super::super::error::{WinError, WinResult};
```
ruabmbua: I am not really a fan of relative paths for imports, when its necessary to go up multiple levels.
Its much clearer with one glance from *where* the item was imported with the absolute path.

src/lib.rs
```diff
@@ -58,48 +58,68 @@
...
 
-#[cfg(hidapi)]
-use crate::hidapi::HidApiBackend;
-#[cfg(all(feature = "linux-native", target_os = "linux"))]
-use linux_native::HidApiBackend;
-#[cfg(all(feature = "windows-native", target_os = "windows"))]
-use windows_native::HidApiBackend;
+cfg_if! {
+    if #[cfg(all(feature = "linux-native", target_os = "linux"))] {
+        //#[cfg_attr(docsrs, doc(cfg(all(feature = "linux-native", target_os = "linux"))))]
```
ruabmbua, line 76: Why comment out these cfg_attrs, and not the ones below?

src/windows_native/descriptor/encoder.rs
sidit77: I made this change to keep the backend more self-contained and not depend on the name it was imported under. My original idea was to do something like this:
```rust
#[cfg_attr(feature = "linux-native", path = "linux_native.rs")]
#[cfg_attr(feature = "windows-native", path = "windows_native/mod.rs")]
mod platform;
```
Which obviously broke those imports.

After changing my approach to `cfg_if` I decided to nevertheless keep it. I can revert the paths if you want.

src/lib.rs
sidit77: The attribute is used to document that a certain function is limited to certain configurations:
image

The native backends should not appear to the public api and therefore the attribute seemed pretty useless. However I just commented them out for now to make it easier to reenable them in case I am wrong and they did actually serve a purpose that I missed.

ruabmbua: Yeah I think you are right. We can just remove them completely, since its backend code.

ruabmbua:
My last concern was, that cfg_if completely breaks IDE support. Apparently its now limited to only being able to index one of multiple cfg_if branches, which should be good enough.

PR #139, opened 2024-01-13
Document what read and write methods return

### LINKED ISSUE #134: Document what write() and read() functions return
What do the `HidDevice::write()` and `HidDevice::read()` functions return when it is successful?

It's a `usize` type but what does it mean? My guess is it's how many bytes were written from or read into the buffer. But would the length of the buffer ever not equal the return value?

Nothing about the return value is documented with the functions.

YgorSouza:
Closes #134

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -534,6 +534,8 @@ impl HidDevice {
         self.inner.check_error()
     }
 
+    /// Write an Output report to a HID device.
+    ///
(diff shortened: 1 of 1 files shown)


ruabmbua (changes requested):
Thank you very much for updating the docs!

ruabmbua (approved):

PR #144, opened 2024-01-26
Reduce package size

uklotzde:
The crate package does not need to include all files. Especially not everything in /etc/hidapi.

#### Before

Packaged 276 files, 2.2MiB (384.3KiB compressed)

#### After

Packaged 65 files, 472.4KiB (116.4KiB compressed)

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -5,6 +5,7 @@ on:
     branches: [ master ]
   pull_request:
     branches: [ master ]
+  workflow_dispatch:
 
 env:
   CARGO_TERM_COLOR: always
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -53,6 +53,10 @@ jobs:
       run: cargo build --no-default-features --features linux-${{ matrix.linkage }}-${{ matrix.library }} --verbose
     - name: Run tests
       run: cargo test --no-default-features --features linux-${{ matrix.linkage }}-${{ matrix.library }} --verbose
+    - name: Verify package
+      run: |
+        rm -f ./etc/libusb/README.md
+        cargo package --no-default-features --features linux-${{ matrix.linkage }}-${{ matrix.library }} --verbose
 
   build-linux-native:
     runs-on: ubuntu-latest
@@ -71,6 +75,8 @@ jobs:
       run: cargo build --no-default-features --features linux-native --verbose
     - name: Run tests
       run: cargo test --no-default-features --features linux-native --verbose
+    - name: Verify package
+      run: cargo package --no-default-features --features linux-native --verbose
 
   build-windows:
     runs-on: windows-latest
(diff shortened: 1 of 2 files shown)


uklotzde:
Unsure about which tasks in the GitHub CI workflow actually need to include the `cargo package` step for verification.

ruabmbua:
Lets try it on CI.

ruabmbua:
Ah I am pretty sure you forgot the --no-default-features flag a few times.

uklotzde:
> Ah I am pretty sure you forgot the --no-default-features flag a few times.

I have force pushed after adding all ` --no-default-features` arguments.

uklotzde:
With the manual dispatch enabled everyone would be able to run the CI workflow on demand on development branches.

uklotzde:
Now it should work: https://github.com/uklotzde/hidapi-rs/actions/runs/7682363490

.github/workflows/rust.yml
```diff
@@ -53,6 +54,10 @@ jobs:
       run: cargo build --no-default-features --features linux-${{ matrix.linkage }}-${{ matrix.library }} --verbose
     - name: Run tests
       run: cargo test --no-default-features --features linux-${{ matrix.linkage }}-${{ matrix.library }} --verbose
+    - name: Verify package
+      run: |
+        rm -f ./etc/libusb/README.md
```
ruabmbua, line 60: Its not entirely clear to me why you have to rm the README here. Can you maybe explain?

ruabmbua (changes requested):

.github/workflows/rust.yml
uklotzde: Looks like a symbolic link to the actual `README` file in the libusb repo. Instead of simply renaming the file they have chosen this unorthodox approach.

It suddenly appears during one of the preceding build steps. Should either be fixed upstream or in this repository.

Passing `--allow-dirty` instead as a wildcard allowance workaround could be dangerous, not recommended.

ruabmbua: Ok, maybe just add a short comment to the CI file to make that clear. After that we can merge the PR.

uklotzde: Why is /etc/hidapi a submodule but /etc/libusb not?

ruabmbua: I just found out myself, apparently its just done in the CI step to install libusb as a dependency. I think the problem was, that the old CI images in github actions had incompatible / old versions. 

Would probably be better to build that stuff in some kind of temporary directory. Not sure if mktemp would work.

uklotzde: Done.

ruabmbua:
Thx for the contribution. To be honest I actually learned something new about cargo this time. I had no idea you could manually control what files were included when it packages source files. I always thought it did something weird specific to VCS system like git (which would be strange), but now I know better.

.github/workflows/rust.yml
uklotzde: Fixed: https://github.com/ruabmbua/hidapi-rs/pull/148/commits/19d0d3dfd4f0d048d2172f5142fd7532c58f63b3

The `etc/libusb/` directory in my local file system was probably a leftover from a previous checkout. Nothing to do here.

PR #150, opened 2024-02-06
hiddevice: add `AsFd` impl

JohannesSchilling:
This is useful to allow performing custom ioctls/operations on nonstandard HID devices.

### DIFF
--- a/src/linux_native.rs
+++ b/src/linux_native.rs
@@ -8,7 +8,7 @@ use std::{
     fs::{File, OpenOptions},
     io::{Cursor, Read, Seek, SeekFrom},
     os::{
-        fd::{AsRawFd, OwnedFd, AsFd, BorrowedFd},
+        fd::{AsFd, AsRawFd, BorrowedFd, OwnedFd},
         unix::{ffi::OsStringExt, fs::OpenOptionsExt},
     },
     path::{Path, PathBuf},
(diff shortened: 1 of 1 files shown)


ruabmbua:
Its probably better to impl https://doc.rust-lang.org/std/os/fd/trait.AsFd.html ?

ruabmbua:
Also, would you mind looking into if we can do the same for the rust windows backend?

JohannesSchilling:
hm, on first look windows looks more difficult, because there we use https://docs.rs/windows-sys/latest/windows_sys/Win32/Foundation/type.HANDLE.html as basis, but the [`AsHandle` trait](https://doc.rust-lang.org/std/os/windows/io/trait.AsHandle.html) uses https://doc.rust-lang.org/std/os/windows/io/type.RawHandle.html, so we'd have to cast `isize` to `*mut c_void` (?)

ruabmbua:
I say lets ignore windows for now. I looked through some comments e.g. https://github.com/microsoft/windows-rs/issues/2222. If windows_sys devs decide to implement something, we can use that.

ruabmbua:
Seems like your changes do not compile, pls check.

ruabmbua:
Thx, will release a new version to crates.io soon.

PR #152, opened 2024-02-23
Use CancelIoEx instead of CancelIo on windows_native

### LINKED ISSUE #151: HidDevice::drop is not thread safe on Windows
I came across this problem in my application and managed to write a minimal reproduce for it:

<details><summary>Click to expand</summary>

```rs
#![forbid(unsafe_code)]
use std::{sync::Mutex, time::Duration};

use hidapi::HidDevice;

fn main() {
    let hidapi = hidapi::HidApi::new().unwrap();
    let devices = Mutex::new(Vec::<HidDevice>::new());
    std::thread::scope(|s| {
        s.spawn(|| {
            let mut buffer = [0; 2000];
            for _ in 0..10_000 {
                for device in devices.lock().unwrap().iter() {
                    _ = device.read(&mut buffer);
                }
                std::thread::sleep(Duration::from_millis(1));
            }
        });
        for _ in 0..2000 {
            {
                let mut devices = devices.lock().unwrap();
                devices.clear();
                devices.extend(hidapi.device_list().map(|d| {
                    let d = d.open_device(&hidapi).unwrap();
                    d.set_blocking_mode(false).unwrap();
                    d
                }));
                assert!(!devices.is_empty());
            }
            std::thread::sleep(Duration::from_millis(1));
        }
    });
}
```
</details>

This simply consists of two threads, where one repeatedly opens and closes all connected devices, and the other tries to read all devices that are open. It only uses safe Rust, and only depends on hidapi 2.6.0, and when I run it with some custom HID devices connected, it crashes almost every time with either `(exit code: 0xc0000374, STATUS_HEAP_CORRUPTION)` or `(exit code: 0xc0000005, STATUS_ACCESS_VIOLATION)`. This happens with or without the `windows-native` feature.

I believe this is due to the fact that the `Drop` implementation calls [CancelIo](https://learn.microsoft.com/en-us/windows/win32/fileio/cancelio), which only cancels I/O operations started on the calling thread. I think it should use [CancelIoEx](https://learn.microsoft.com/en-us/windows/win32/fileio/cancelioex-func) instead.
With the `windows-native` feature enabled and the following patch, it did not crash during my 30+ tests:

<details><summary>Click to expand</summary>

```diff
diff --git a/src/windows_native/mod.rs b/src/windows_native/mod.rs
index cffbca2..0726fc4 100644
--- a/src/windows_native/mod.rs
+++ b/src/windows_native/mod.rs
@@ -46,7 +46,7 @@ use windows_sys::Win32::Storage::FileSystem::{
     OPEN_EXISTING,
 };
 use windows_sys::Win32::System::Threading::ResetEvent;
-use windows_sys::Win32::System::IO::{CancelIo, DeviceIoControl};
+use windows_sys::Win32::System::IO::{CancelIo, CancelIoEx, DeviceIoControl};

 const STRING_BUF_LEN: usize = 128;

@@ -335,7 +335,7 @@ impl HidDeviceBackendWindows for HidDevice {
 impl Drop for HidDevice {
     fn drop(&mut self) {
         unsafe {
-            CancelIo(self.device_handle.as_raw());
+            CancelIoEx(self.device_handle.as_raw(), null());
         }
     }
 }
```
</details>

This has been a topic of discussion in the original C library for many years (see for example  <https://github.com/libusb/hidapi/issues/133> and <https://github.com/signal11/hidapi/issues/48>), but I believe it is the correct solution at least for the Rust crate. I wasn't able to test if the [other call](https://github.com/ruabmbua/hidapi-rs/blob/c2b41a517b419af45666afa92fb9c450d8e80c6c/src/windows_native/mod.rs#L182) to `CancelIo` needs to be changed to `CancelIoEx` as well, since my application only reads and writes from one thread.

For completeness, although I don't think it matters, I ran these tests on Windows 11 Pro, Version 23H2.

YgorSouza:
Fixes a use-after-free bug when dropping the HidDevice from a thread other than the one that was last used to read it, because of a pending async operation that did not get canceled.

If the cancellation does not return an error, we get the overlapped result to block until it actually goes through. This is recommended by Microsoft in this example:
<https://learn.microsoft.com/en-us/windows/win32/fileio/canceling-pending-i-o-operations#canceling-asynchronous-io>.

Closes #151

### DIFF
--- a/src/windows_native/mod.rs
+++ b/src/windows_native/mod.rs
@@ -46,7 +46,7 @@ use windows_sys::Win32::Storage::FileSystem::{
     OPEN_EXISTING,
 };
 use windows_sys::Win32::System::Threading::ResetEvent;
-use windows_sys::Win32::System::IO::{CancelIo, DeviceIoControl};
+use windows_sys::Win32::System::IO::{CancelIoEx, DeviceIoControl};
 
 const STRING_BUF_LEN: usize = 128;
 
@@ -179,7 +179,7 @@ impl HidDeviceBackendBase for HidDevice {
             if res != TRUE {
                 let err = Win32Error::last();
                 if err != Win32Error::IoPending {
-                    unsafe { CancelIo(self.device_handle.as_raw()) };
+                    unsafe { CancelIoEx(self.device_handle.as_raw(), state.overlapped.as_raw()) };
                     self.read_pending.set(false);
                     return Err(err.into());
                 }
@@ -335,7 +335,16 @@ impl HidDeviceBackendWindows for HidDevice {
 impl Drop for HidDevice {
     fn drop(&mut self) {
         unsafe {
-            CancelIo(self.device_handle.as_raw());
+            for state in [
+                &mut self.read_state,
+                &mut self.write_state,
+                &mut self.feature_state,
+            ] {
+                let mut state = state.borrow_mut();
+                if CancelIoEx(self.device_handle.as_raw(), state.overlapped.as_raw()) > 0 {
+                    _ = state.overlapped.get_result(&self.device_handle, None);
+                }
+            }
         }
     }
 }


ruabmbua:
Thanks for fixing this, will release a new version to crates.io.

PR #158, opened 2024-07-04
Add feature for doc_cfg, specify allowed cfgs in build.rs

Tiwalun:
Specify possible configs as described in <https://doc.rust-lang.org/nightly/rustc/check-cfg.html>, and fix compilation with `--cfg docsrs`.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -25,6 +25,9 @@ use std::env;
 fn main() {
     let target = env::var("TARGET").unwrap();
 
+    println!("cargo::rustc-check-cfg=cfg(hidapi)");
+    println!("cargo::rustc-check-cfg=cfg(libusb)");
+
     if target.contains("linux") {
         compile_linux();
     } else if target.contains("windows") {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -59,6 +59,7 @@
 //! Since `hidapi` 0.12 it is possible to open MacOS devices with shared access, so that multiple
 //! [`HidDevice`] handles can access the same physical device. For backward compatibility this is
 //! an opt-in that can be enabled with the `macos-shared-device` feature flag.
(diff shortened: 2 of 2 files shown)


JohnAZoidberg:
Seems this is only supported in nightly :/

```
> cargo build
    Updating crates.io index
  Downloaded hidapi v2.6.2
  Downloaded 1 crate (128.1 KB) in 0.26s
   Compiling libusb1-sys v0.7.0
   Compiling hidapi v2.6.2
error: unsupported output in build script of `hidapi v2.6.2`: `cargo::rustc-check-cfg=cfg(hidapi)`
Found a `cargo::key=value` build directive which is reserved for future use.
Either change the directive to `cargo:key=value` syntax (note the single `:`) or upgrade your version of Rust.
See https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script for more information about build script outputs.
warning: build failed, waiting for other jobs to finish...

> cargo --version
cargo 1.75.0 (1d8b05cdd 2023-11-20)
```

Tiwalun:
I think this is related to the use of two colons in `cargo::rustc-check-cfg=cfg(hidapi)`, if compatibility with versions older than 1.77 is required that has to be changed to `cargo:rustc-check-cfg=cfg(hidapi)`. I don't think the actual `rustc-check-cfg` directive should cause a problem.

@ruabmbua Does this crate have any policies regarding which rust versions are supported?

ruabmbua:
There is no explicit MSRV, but I try to keep it working for some (reasonably) recent packaged rustc versions from linux distros. Seems like debian stable does not have 1.77+ yet, so I will make a new release with a fix.

ruabmbua:
Fixed in https://github.com/ruabmbua/hidapi-rs/releases/tag/v2.6.3

PR #159, opened 2024-09-23
set_output_report

ArtyomPrisyazhnyy:


ruabmbua:
Any interest in still working on this? Right now it does not build for any of our platforms in CI.

marek-g:
Hi! I am interested in this as I need it for Redragon K585 keyboard.

The problem is that `hid_set_output_report` is available in unreleased yet `hidapi`'s version 0.15. As I understand CI uses the latest version available on Ubuntu, which is currently v0.14. The version from github is needed.

Probably I will be able to hack the solution locally for myself, but I don't know how to configure it here in CI so everybody is happy (how to fetch source code from github? hide it under some crate feature, so it compiles for users of v0.14 and v0.15? wait for 0.15 to be released?).

PR #161, opened 2024-11-10
linux-native: provide better signatures for the ioctl wrappers

carlosmn:
It looks like I missed the write-only or read-only versions of these macros which let us specify that a ioctl is only writing or only reading.

Use them so we match the actual functionality. This also allows us to remove a copy when sending a feature report as we don't have to work around the misidentified mutability.


ruabmbua:
Thx!

PR #167, opened 2025-03-09
Add `linux-native-basic-udev` feature

xobs:
Add a new feature, `linux-native-basic-udev`. This is identical to `linux-native` but uses `basic-udev` as a udev source rather than `udev-rs`.

[basic-udev](https://github.com/xobs/basic-udev) is a reimplementation of just enough udev functionality to support device discovery on most common platforms without needing to link libudev.so.

As a result, it becomes possible to cross-compile crates that depend on `hidapi-rs` to Linux (both GNU and Musl) without needing a full sysroot. This greatly simplifies cross-compiling to multiple target architectures, particularly when using cargo-zigbuild.


Cargo.toml
```diff
@@ -48,6 +48,7 @@ linux-static-hidraw = []
 linux-shared-libusb = []
 linux-shared-hidraw = []
 linux-native = ["dep:udev", "dep:nix"]
+linux-native-basic-udev = ["dep:basic-udev", "dep:nix"]
```
ruabmbua, line 51: I think its not necessary to create a `backend` feature for just swapping out a compatible alternative implementation of a dependency. Consider just creating a udev and basic-udev feature and selecting the normal udev feature by default.

README.md
```diff
@@ -42,5 +42,13 @@ let res = device.write(&buf).unwrap();
 println!("Wrote: {:?} byte(s)", res);
 ```
 
+# Cross-compiling
```
ruabmbua: Not sure if its a great idea to include documentation for cross compiling, maybe cut it down to just mentioning the relevant feature flags of the crate that help making it easier?

ruabmbua (commented):
Thanks for working on this. If you want I can make small fixes myself after merging, or you can do it yourself.

Cargo.toml
xobs: How would that work? Wouldn't that require people who select `linux-native` to also select the new `udev` feature? Or that would force users to all opt-in to the new `udev` crate, even those who are compiling for e.g. illumos? My concern is that as long as you select the `udev` crate, it will run the pkgconfig script at build time, which may impact users who don't want to depend on udev.

README.md
xobs: Sure, I can do that.

Cargo.toml
ruabmbua: Ah right, my mistake. Then lets keep it like it is!

On another note maybe you can add the new linux-native-basic-udev feature as one of the CI  test targets?

xobs: Sure. I added it to CI. I added both an Ubuntu target as well as one that cross-compiles to arm-musleabihf from Windows just to make sure that it does function as expected.

xobs: ...right. `cargo test` probably isn't a good idea to run when cross-compiling.

PR #176, opened 2025-12-19
Clean up some lints

carlosmn:
I was looking around the code and saw that clippy has some suggestions here. Mostly it's about being less verbose.

I haven't tested the co2mon example as I'm away from my device that talks that protocol, but it doesn't seem like it should break anything and I did test out the equivalence in the playground.


ruabmbua:
thanks!

PR #177, opened 2025-12-19
linux-native: do not panic on invalid strings from the OS

### LINKED ISSUE #166: Panic at expect "udev strings should always be utf8"
Hello! I'm not completely sure, but seems I have some strange device plugged in that breaks the assumption that udev strings must be utf8. I'm using probe-rs, which becomes unusable because of this.

Here's the backtrace:
```
thread 'main' panicked at /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/hidapi-2.6.3/src/linux_native.rs:386:19:
udev strings should always be utf8
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: hidapi::linux_native::device_to_hid_device_info
   3: <core::iter::adapters::filter::Filter<I,P> as core::iter::traits::iterator::Iterator>::next
   4: <alloc::vec::Vec<T> as alloc::vec::spec_from_iter::SpecFromIter<T,I>>::from_iter
   5: hidapi::linux_native::HidApiBackend::get_hid_device_info_vector
   6: hidapi::HidApi::add_devices
   7: hidapi::HidApi::new
   ...
```

Can provide more details if needed :)

carlosmn:
While strings that come from udev should indeed be UTF-8 we cannot expect that user-provided fields contain anything valid. Make a similar conversion to the C hidapi version when we encounter an invalid string.

Fields like e.g. the serial are device or user-controlled and there are UHID libraries out there which fill it in with random values.

This should fix #166. I don't know if it's too much effort for nonsense values to try to keep the unicode bits working but that's what the C hidapi version does (though it does start with wchar and I haven't tested what happens if you put an emoji in your serial).

### DIFF
--- a/src/linux_native.rs
+++ b/src/linux_native.rs
@@ -17,6 +17,7 @@ use std::{
     path::{Path, PathBuf},
 };
 
+use libc::wchar_t;
 use nix::{
     errno::Errno,
     poll::{poll, PollFd, PollFlags},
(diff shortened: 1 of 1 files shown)


ruabmbua:
I agree, its probably ok to do the conversion just like this.

Thanks!

Issue #51, opened 2020-02-13
Async interface


NickeZ:
I made an [async interface to this crate](https://github.com/digitalbitbox/bitbox-bridge/tree/master/hidapi-async). However, since there is no way to register to become awakened when there is data to read I have to create a thread and do blocking reads.

Reading through libhidapi (the C library) I realised that the libusb interface actually is asynchronous and that libhidapi conveniently (unfortunately) wraps away that using an internal thread. So the result feels incredible clunky with two threads and multiple mutexes / synchronization primitives.

I think the best approach actually is to skip hidapi and rewrite using the platform specific bindings from scratch. But before I start such a big effort I wanted to check in here if anyone else already is working on it.

Thanks

ruabmbua:
Hi @NickeZ .

I was already working on a raw rust backend for this crate, which would skip hidapi (the C library), and use the os interfaces directly.

At the time I did this, I also wanted to add async support, but the rust async story was not really stable. I decided to wait, until async ecosystem / language integration catches up. I have not looked into async / await since then, but from what I heard, it might be ready

Regarding libusb, I do not plan on using it at all. All somewhat modern Linux systems should have the hidraw kernel interface.

I might be able to spend some time on this starting next week, but I have some other hobby projects going on.

https://github.com/ruabmbua/hidapi-rs/projects/1

ruabmbua:
If you have any pointers, if I should try it with the tokio ecosystem, or async-std, please tell me.
Maybe it`s possible to be runtime independent by just using mio, std::future and std::task?

NickeZ:
I haven't looked close enough to know exactly what crates are needed, but I think you shouldn't need `tokio`. I think technically it should be enough with `std::future`, `std::task` and `AsyncRead/Write` from `futures`. However it would be great if we could reuse some platform specific code from `mio` to avoid reimplementing it. But it sounds to me like that depends on how usb works on the different platforms.

One caveat is that we may have to implement `AsyncRead/Write` from `tokio` as well to support that ecosystem. Currently it is in a bit of flux how the final std traits will look.

lnicola:
Tokio will migrate to the version of the async traits that gets standardized, but until then it would be nice to support both, if possible.

NickeZ:
So the plan would be to wrap these interfaces?

* Windows (using hid.dll)
* Linux (using the Kernel's hidraw driver)
* Mac (using IOHidManager)

brokenthorn:
[...]

PR #178, opened 2026-01-01
Add async for linux-native

carlosmn:
This goes some of the way towards taking care of #51 though right now just for linux-native as it's the easiest to do (at least for me).

Doing async through traits can be a bit challenging. I played around with methods returning boxed futures but then you start having to care about the lifetimes of the buffers and those go into the signatures and it gets ugly.

Which is why I went with how everyone else does it, which is to have a `poll_{read,write}` method. We then use `futures::poll_fn` to make a future out of that.

As far as choosing which runtime flavour to use (given you can't go fully generic without implementing a bunch more things that the runtimes give you), I ended up going for both major ones as it was pretty easy to implement them both. The tokio variant works if you're using tokio without extra dependencies, and the smol/async-io version works wherever.

Unfortunately this is quite an explosion of feature flags to enable the user of the library to say what stack they're working on, but I don't know of a way to auto-select depending on what else is selected in a dependent project. And I haven't even tackled what happens if you want async but also basic-udev.

One thing I'm not sure here is the naming of the generic `async` feature as you're not meant to enable it yourself, but rather choose a backend. Maybe it should be `__async` to indicate it's internal. This is e.g. what I've seen `reqwest` does.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,6 +43,7 @@ include = [
 
 [features]
 default = ["linux-static-hidraw", "illumos-static-libusb"]
+async = ["dep:futures"]
 linux-static-libusb = []
 linux-static-hidraw = []
 linux-shared-libusb = []
@@ -67,6 +68,7 @@ windows-native = [
 [dependencies]
 libc = "0.2"
 cfg-if = "1"
+futures = { version = "0.3", optional = true }
 
 [target.'cfg(target_os = "linux")'.dependencies]
 udev = { version = "0.8", optional = true }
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -64,9 +64,21 @@ windows-native = [
...
+
+# Enables some common code to do with async
+__async = ["dep:futures"]
+
 [dependencies]
 libc = "0.2"
 cfg-if = "1"
+futures = { version = "0.3", optional = true }
+async-io = { version = "2", optional = true }
+tokio = { version = "1", optional = true, features = ["net"] }
```
ruabmbua: Why do we need the 'net' feature?

PR #179, opened 2026-01-16
Upgrade dependencies, in particular `windows-sys` to 0.61 and `nix` to 0.30

ColinFinck:
This PR is the result of `cargo upgrade --incompatible --pinned --ignore-rust-version --recursive` and manual modifications to account for the changes in the upstream crates.

`windows-sys` 0.61.x is very important for me, because it consistently uses `raw-dylib` for importing APIs from Windows DLLs. This makes a huge difference when using non-MSVC toolchains to build Windows binaries, e.g. `x86_64-pc-windows-gnullvm`.

Unfortunately, `Win32_Storage_EnhancedStorage` is only available in the heavyweight `windows` crate these days, which is why I also need to import that one. However, I'm only importing it for that one feature and continue to use `windows-sys` for the rest of the code.

CC @ruabmbua

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,11 +53,11 @@ illumos-static-libusb = []
 illumos-shared-libusb = []
 macos-shared-device = []
 windows-native = [
+    "windows/Win32_Storage_EnhancedStorage",
     "windows-sys/Win32_Devices_DeviceAndDriverInstallation",
     "windows-sys/Win32_Devices_HumanInterfaceDevice",
     "windows-sys/Win32_Devices_Properties",
     "windows-sys/Win32_Security",
-    "windows-sys/Win32_Storage_EnhancedStorage",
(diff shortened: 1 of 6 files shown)


ColinFinck:
Bump

CC @ruabmbua

ruabmbua:
Hey, sry for forgetting about this pr.

regarding the new dependency, it would probably be better to try adding the missing extern function to windows-sys?

we can temporarily add the extern directly to this crate until it’s in windows-sys.

I do not want to depend on a new crate for one missing extern function

ColinFinck:
@ruabmbua Thank you for your feedback. I've addressed your concerns in 722f0072a71c0820d9def912d79491d075fd31e9. Was surprisingly easy to get rid of the heavyweight `windows` crate dependency again :)

PR #182, opened 2026-04-24
linux-native: Change ioctls back to readwrite, to match kernel definitions

Mstrodl:
Here are the definitions from the kernel:
```c
#define HIDIOCSFEATURE(len)    _IOC(_IOC_WRITE|_IOC_READ, 'H', 0x06, len)
#define HIDIOCGFEATURE(len)    _IOC(_IOC_WRITE|_IOC_READ, 'H', 0x07, len)
#define HIDIOCSOUTPUT(len)    _IOC(_IOC_WRITE|_IOC_READ, 'H', 0x0B, len)
#define HIDIOCGOUTPUT(len)    _IOC(_IOC_WRITE|_IOC_READ, 'H', 0x0C, len)
```

Using read and write instead of readwrite meant we were trying to use an invalid ioctl number, causing the ioctl to return EINVAL (at least on my system...). I should note I have only tested `HIDIOCSFEATURE`, though.

Happy to make this one commit instead of a revert and a commit to fix the other features that have been added since (outputs) if that's preferred.

### DIFF
--- a/src/linux_native.rs
+++ b/src/linux_native.rs
@@ -536,7 +536,14 @@ impl HidDeviceBackendBase for HidDevice {
             return Err(HidError::InvalidZeroSizeData);
         }
 
-        let res = match unsafe { hidraw_ioc_set_feature(self.fd.as_raw_fd(), data) } {
+        // Have to crate owned buffer, because its not safe to cast shared
+        // reference to mutable reference, even if the underlying function never
+        // tries to mutate it.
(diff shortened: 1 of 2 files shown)


deldotbrain:
I'm seeing the same `EINVAL`, and these changes fix the problem for me as well.

ruabmbua:
Thx, seems like this got messed up sometime ago!

PR #184, opened 2026-07-29
Fix empty HID_UNIQ causing devices to be dropped from enumeration in linux-native

dainnilsson:
When a HID device has en empty HID_UNIQ, libudev returns NULL, causing property_value() to return None. This in turn causes the device to be dropped from enumeration. This happens with eg. YubiKeys which do not expose a serial number in the USB descriptor. The proposed fix is to instead use an empty string for these devices.


ruabmbua:
Can you check if the current behavior is a mismatch with the older hidapi C library based hidraw implementation on linux?

If it is, and it also returns an empty string instead, comment that fact in your code, or document it in the commit

dainnilsson:
I've clarified the comment (and commit message) with comparisons to the C implementation. I have also locally tested enumeration using both backends, and confirmed that the native backend with this change matches the behavior of the C one.

ruabmbua:
Thanks, and sorry this completely slipped my mind!

